PR #3, opened 2019-03-08
Added bioconductor fields and package coverage

dgkf:
### Changes
* added `emmeans` code coverage
* used `BiocPkgTools::biocPkgList` to get analogous fields as CRAN db

### Needs work:
* haven't yet re-generated dataset for package, wasn't sure how we wanted to version this on GitHub
* `Copyright` isn't an avaialble field from `biocPkgList`
* `limma` doesn't use a testing framework I'm familiar with, not sure how to test it

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -1,0 +1,5 @@
+^codecov\.yml$
+^appveyor\.yml$
+^\.travis\.yml$
+^.*\.Rproj$
+^\.Rproj\.user$
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -8,11 +8,14 @@ Description:
 License: What license is it under?
 Encoding: UTF-8
 LazyData: true
+biocViews:
 Depends:
     dplyr,
     purrr,
     covr,
-    tools
+    tools,
+Imports: 
+    BiocPkgTools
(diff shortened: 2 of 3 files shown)


elong0527:
@dgkf :

1. how would you suggest for the number of downloads derivation? 

Would you think the approach used by `dlstats` package is OK? 

https://cran.r-project.org/web/packages/dlstats/vignettes/dlstats.html

2. Is there a way to extract package dependency from `limma` 

3. Would you think we should clean Author to a listing as we did for dependency? (low priority)

dgkf:
@elong0527 I'd opt to use the `dlstats` package. Looks like it can handle both CRAN and bioconductor stats. Just to collect the info in one place, I also discovered the function `BiocPkgTools::biocDownloadStats()`, although it looks like it exposes just about the same information. Is there any reason why you might feel that `dlstats` wouldn't suit our needs?

Regarding the other two bullets, I added bionconductor dependencies and cleaned the Author column. 

Let's try to get this new body of work pulled into master before expanding the scope any more than it already is. I'd prefer not to let this pull request get too far ahead of master. I've added @elong0527 for review, so if everything looks reasonable, feel free to approve and pull it in.

PR #9, opened 2019-03-26
Create proof_of_concept.md

elong0527:
Draft proof of concept for risk metrics

### DIFF
--- a/reports/proof_of_concept.md
+++ b/reports/proof_of_concept.md
@@ -1,0 +1,35 @@
+# Proof of Concept for Risk Metrics
+
+## Goal
+
+* Overall Goal: An automatically pipline to create metrics for R packages on CRAN/Bioconductor
+
+## Scope 
+
+* Create a database that contain the risk metrics for every R package of each version.
+* Define risk metrics
+* Derive code coverage 
+
+### Out-of-Scope
+
+* Define the risk score based on risk metrics (pending confirmation)
+
+## Achievement as of 03/31/2019
+
+
+* Create a [function to collected risk metrics](https://github.com/pharmaR/riskmetric/blob/master/R/riskmetric.R) for five R packages as pilot
+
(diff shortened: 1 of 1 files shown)


reports/proof_of_concept.md
```diff
@@ -0,0 +1,35 @@
+# Proof of Concept for Risk Metrics
+
+## Goal
+
+* Overall Goal: An automatically pipline to create metrics for R packages on CRAN/Bioconductor
+
+## Scope 
+
+* Create a database that contain the risk metrics for every R package of each version.
```
dgkf: Should we discuss a scope for packages to consider? Perhaps target ~1000 packages representative of a core cohort first (e.g. it would be okay to have incomplete metrics or fail for some packages as long as we can hit a representative cohort initially)

elong0527: Yeah, I like your suggestion. Cover the exponentially growing R packages on CRAN could be hard to achieve at current stage. 

How about all R packages with unit test (using testthat flow on CRAN, to ensure we can derive code coverage for them) + some specifically identified R packages (may need extra works to gather metrics) by the working group (e.g. emmeans need to gather information from both CRAN and github repo)

matthiazzz: Do you guys already use R packages in productive environment? How about we start collecting those as a base list? 

I could start with about 150 packages (including dependencies) as a starting point?

Also, we could check on commercial products like ValidR and see what packages they have "validated"?

dgkf: This sounds like a really good way of narrowing the scope in a way that is still high priority to all of our organizations. Perhaps we could start a shared doc here where we can collect packages and sources? I don't think it needs to be anything super rigorous for the long term, but we could at least collect it all in one place.

I'm thinking a csv of the format

|package|source|
|---|---|
|dplyr|CRAN|
|limma|Bioconductor|
|gt|https://github.com/rstudio/gt|

elong0527: Sure @dgkf , that is great. Would you want to initiate a file in the data folder? 

Let me close this pull request. And we can discuss the package listing in its specific place.

PR #10, opened 2019-04-08
+ pkgdown + testthat + MIT + reduced to single CRAN version example

kkmann:
Hey, 
this is a major overhaul of the basic infrastructure. I added testthat infrastructure, automatic code coverage reports (needs set up by repository owner), automatic pkgdown deployment (needs setup, see here https://pkgdown.r-lib.org/reference/deploy_site_github.html) and reduced the example to a bare minimum to get rid of the bioconductor dependencies.
This still needs some tinkering before the CI/CD + codecov + pkgdown works smoothly, @elong0527 can you set that up at some point?


kkmann:
sorry for messing up the master branch before ;) you might want to set up a branch protection rule there ;)

tests/testthat/test_cran_metrics.R
```diff
@@ -0,0 +1,21 @@
+context("Test CRAN related metrics")
+
+test_that("CRAN current version returns correct format", {
+
+    expect_equal(
+      grep(
+        "[0-9]+.[0-9]+.[0-9]+(.[0-9]+)*",
```
dgkf: Be careful with the `.` usage here - `.` is a reserved character in `regex` and will capture any characters such that "1a2b3cd" would be a valid version number. Not a huge deal, and the standard version format will still get picked up. 

Consider using `"\\d+(?:\\.\\d+){2,}"`

R/memoise_cran_db.R
```diff
@@ -0,0 +1,22 @@
+# taken from https://github.com/ropenscilabs/packagemetrics/blob/master/R/get_cran.R
+memoise_cran_db <- memoise::memoise({
```
dgkf: Super cool. I'll be using this! I always write an annoying amount of boilerplate to handle this type of thing.

tests/testthat/test_cran_metrics.R
kkmann: oh dear, regular expressions certainly are not my strong suit ;) thx for pointing that out!

R/memoise_cran_db.R
kkmann: functional programming roxxx ;)

kkmann:
@elong0527 since this is almost a complete overhaul, I'll wait for your opinion as well. can you set up all the ci/cd stuff after merging?

elong0527:
@kkmann , I am waiting for the permission approve from @andyofsmeg to deploy the pkgdown. Will complete the set up after approval.

PR #12, opened 2019-04-15
Added handling of get_cran_version for vector of package names

dgkf:
I figured if we're operating on the `cran_db` data.frame that it would be good to be able to accommodate a situation where we want to pull a whole vector of version numbers. 

I also updated the test suite with a couple more tests to accommodate the new behavior.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,8 @@
 Package: riskmetric
 Type: Package
 Title: Risk Metrics to Evaluatin R Packages
-Description: todo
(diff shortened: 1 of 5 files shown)


dgkf:
Ah, thanks for pointing that out, @kkmann. I was out of the loop for a few days and jumping back in. Guess I missed (or more likely, forgot) some of the recent conversations. 

I'll complete the pull request because I think I made some positive increments, though you're right that it's best to take a step back and think more critically about structure before doing too much more development.

PR #17, opened 2019-05-03
added proposal for an S3 class to store package source reference 

dgkf:
I put together a proof-of-concept of what an S3 class might look like for our package reference info. It currently contains minimal data, but could be used to dispatch to other functions for aggregating package metadata needed to provide a risk score.

The plan here is to create a single function that can be called with a variety of ways of specifying a package and provide back an S3 class object that can be dispatched to handle that package source appropriately.

Given `as.pkg_ref(x)` where

|working directory|`x = `| returns output of `class`|
|---|---|---|
|-|`"dplyr"`|`c("pkg_install", "pkg_ref")`(if dplyr installed)|
|riskmetric directory|`"../riskmetric"`|`c("pkg_source", "pkg_ref")`|
|-|`"blah.blah.blah"`|`c("pkg_missing", "pkg_ref")`|
|-|`"dplyr"`|`c("pkg_remote", "pkg_ref")` (if dplyr not installed)|

I think that this approach could help us streamline how we unify methods where possible, and how we handle separate sources elegantly.

### Examples of situations where S3 is beneficial
- If there are some metrics which require source code, we can create functions `as.pkg_source.pkg_install` and `as.pkg_source.pkg_remote` to pull the `URL` field from the description to download the source and return a `pkg_source` reference to the local clone. 
- If we want to assess unit test coverage differently for a remote reference on CRAN and source code cloned locally, we could have two separate functions `get_pkg_coverage.pkg_remote` which might try to pull the coverage out of a badge or CI tool and `get_pkg_coverage.pkg_source` which might run all the tests itself.

### Some areas for further improvement:
- perhaps a global variable such as `options(pkg_refs = c("remote"))` could be used to only consider package references from certain source types, allowing for the metric calculations to run quickly by only scraping data from the web or  alternatively use `options(pkg_refs = c("source", "install", "remotes"))` to always use the most rigorous method available.

Let me know if this sounds like a viable path to take, or if there are glaring gaps in this concept of a package design.


elong0527:
If we go this direction, are we creating additional functions (e.g. `get_xxx_features`) for different methods to get the metric? 

Could we use `dplyr` as an example to illustrate what the output will looks like?

@kkmann , how would you think?

dgkf:
Yes, each `get_xxx` function could be dispatched based on the package reference info we have on hand. For example, if we wanted to assess the test coverage, we might have these functions:

```r
get_test_coverage <- function(x, ...) {
  UseMethod("get_test_coverage")
}

get_test_coverage.default <- function(x, ...) {
  NULL
}

get_test_coverage.character <- function(x, ...) {
  get_test_coverage(as.pkg_ref(x))
}

get_test_coverage.pkg_remote <- function(x, ...) {
  # attempt to scrape test coverage from remote badges, 
  # coveralls.io, codecov.io
}

get_test_coverage.pkg_install <- function(x, ...) {
  # unless explicitly installed, the tests folder won't be included, 
  # default to remote behavior
  if (<installed package contains tests directory>) ...
  else get_test_coverage(as.pkg_remote(x))
}

get_test_coverage.pkg_source <- function(x, ...) {
  covr::package_coverage(x$path)
}
```

Once we have all these defined, we could do `get_test_coverage("dplyr")` to run tests with whatever package data we have available. Or, more practically, we could create a `pkg_ref` object first with our preferred reference type (for example, only building from remote info), then running `get_test_coverage(<pkg_ref_object>)`. 

I think that this style would work well since we could start small by implementing as many tests as possible using only the `*.pkg_remote` dispatched methods before we worry about having to manage package installs or source code, but could easily expand into more robust tests once we're ready just by adding handlers for those classes of package references (`*.pkg_install`, `*.pkg_source`)

elong0527:
Agree! Could you add the get_xxx part in the code and we can merge this pull.

Which part would you think I can work?

dgkf:
Discussed on the phone with @elong0527 and had some comments...
---
* `pkg_ref` object should also have ability to articulate package version specification (package version, git ccommit, etc)
* @dgkf will put together proof-of-concept implementation one metric for a couple formats in a follow-up pull request, but would prefer to merge existing work as to not produce too much work in a single pull request. 
* @kkmann - any additional thoughts on this style of breaking down the metrics assessment?

kkmann:
Love it! A few minor comments:

* naming: it is generally considered bad style now to use dots in R function/method names due to the way R handles S3 method dispatch. I know it looks R-ish but we should really avoid it. I'd suggest renaming this to `pkg_reference` (we do not need the 'as', think constructors and I'd also avoid non-canonic abbreviations wherever possible ;))
* Personally, I am a big fan of NSE; would be happy to put together a PR to enable `pkg_reference(dplyr)` (no ") as well
* I'd also drop the 'get' for sake of brevity for the other methods ;)

dgkf:
Thanks for taking a look @kkmann 

* Agreed on removing the `as.` prefix... that feels like a silly oversight now that you point it out. I'm usually pretty sensitive to not stepping on the S3 toes, but this one slipped past me. 
* I'm a fan of NSE in user-facing code, but since I see this constructor as being predominately for programming usage, I'd lean toward consistency and easiness to reason about. Especially since we will eventually want to iterate over many packages, the most likely use case is to call `pkg_ref` with a variable name not intended to be a package name.
* Agreed, I'm not crazy about the `get` prefix, but I think that some sort of prefix is nice since it helps to informally namespace a collection of related functions whose formal arguments will be consistent. Since this subset of functions will be expected to all take in a package reference and output some sort of standardized output, it's nice for the sake of tab completion discoverability that they all share a prefix (similar to `testthat::test_*`, `stringr::str_*` or `crayon::col_*`). We can probably come up with something more meaningful than "get", though.

dgkf:
This has me thinking a bit more about how this class will be used further downstream. I had a few more thoughts, but decided to break them off as separate issues which you can find in #19 & #20. 

Please check those out, @kkmann. I would love to get your impressions on how these package objects get consumed and used to produce scores that can be composited into a 'package risk assessment' .

PR #23, opened 2019-05-29
[WIP] local package metrics

kkmann:
Okay, a first attempt at trying to compute package metrics locally. I tried to get covr to work but that seems to be a bit involved. I cannot get it to output any tracing info. This is unfortunate since it is a core metric (I presume). Certainly doable, but this requires more digging...

---
*Not ready to merge!*

Updates:

* okay I took an invalid shortcut by running covr on the installed packages directly: covr requires the sources to work properly even if you install the tests. A bit more hazzle but certainly doable. 
* working in principle - need to make check() happy and resolve test errors on travis


kkmann:
hm, okay so I am no thinking: 

1) download package sources as .tar.gz in tmp folder
2) [optionally] install package dependencies to a local temporary library (dependencies should be optional to test within the current ecosystem)
3) install the package in temporary library (that should be our first 'metric')
4) continue with other metrics, covr etc..

kkmann:
I think the best way forward for me is to hijack the devtools internal methods for downloading sources - they are not exported but still better than reinventing the wheel...
I think this is definitely doable and could be pretty solid - but lots of work to do x)

kkmann:
@dgkf I think I managed to get a crude prototype to work. Downloading pkg tarballs of specific version for CRAN and GitHub is already working. Covr also works fine now (still need to make sense of the detailed report x)).

Agreed, I think we can build on this - making the download / temporary library infrastructure a bit more robust over time and then adding a few metrics.

@elong0527 @dgkf do we agree, that we build metrics around a locally available .tar.gz (i.e., parse as much as possible from the DESCRIPTION file itself, like, github url etc.)? Would probably be a good idea to make sure that we have one single temporary library per session... but that is another issue.

kkmann:
jeez what a mess - I'll get back to it another time x)

PR #25, opened 2019-05-29
added get/set functions to improve environment namespace maintainability

dgkf:
added function `new_pkg_env` to instantiate a new package environment, this environment is locked and contains:
* prepopulated variables based on the `pkg_env_fields` list, so that any changes to what variables we should expect in the namespace can be declared in one place
* a `get` function, which validates against the existing variable names to make sure you're requesting a viable variable, can take a `default` argument to fall back to if a variable hasn't been set yet.
* a `set` function which validates against the existing variable names before unlocking the namespace, assigning the new value and re-locking. 
* a `has` function to check if a variable has a value.

With these guardrails in place, I feel a lot more comfortable about using namespaces to accumulate package metadata and mitigate repeated calls to access it.

### DIFF
--- a/R/get_metrics.R
+++ b/R/get_metrics.R
@@ -1,0 +1,23 @@
+
+get_metrics <- function(pkgs) {
+  if (!exists(pkgs))
+    pkgs <- c(
+      get_cran_packages()
+      #get_bioconductor_packages()
+    )
+
+}
+
+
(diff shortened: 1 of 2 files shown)


dgkf:
Closing this pull request as this feature was reworked by overloading the extraction operators and using S3 generics for caching metadata.

Issue #27, opened 2019-07-09
function naming conventions & package structure (part 2)


dgkf:
A continuation of the issue thread #16 now that we've started to settle on a more concrete structure for the package.

In response to @kkmann's suggestions in that thread, I think that having some sort of prefix for the `get_*`-style functions is necessary to keep them easily discoverable. Since all of these functions are inherently similar in their accepted input datatypes and output datatypes, I think that this prefix is valuable for communicating this consistency (similar to `testthat::expect_*` or `ggplot2::geom_*`). 

## naming convention proposal

I've been toying around with the following naming convention and flow for how a summarized score might eventually be developed. It uses three templates for functions, `assess_*` (replacing `get_*`), `score` and `summarize_risk`:

- `assess_*` (e.g. `assess_coverage`, `assess_has_news`)
   This is used for initially gathering related metadata pertaining to an individual metric. This family of functions would be used to transform a package reference into a package metric.
   
   |input|function|output|
   |---|---|--|
   |**S3** `c("pkg_install", "pkg_reference")`|`assess_coverage.pkg_install`|**S3** `c("pkg_metric_coverage", "pkg_metric")`|

- `score` (e.g. `score.pkg_coverage`, `score.pkg_has_news`)
   Ideally this is just a single function that dispatches based on the type of metric that is to be scored such that we can define a conversion from metric metadata to some numeric indication of quality (I propose a 0-1 scale for all scores). For example, for `pkg_has_news`, any presence of a valid NEWS file might be considered a 1 on a 0 to 1 scale. We can then iterate on this if we want more granularity.

   |input|function|output|
   |---|---|---|
   |**S3** `c("pkg_metric_coverage", "pkg_metric")`|`score.pkg_metric_coverage`|**numeric** (0-1)|

- `summarize_risk`
   Finally, with one score per metric we can define a summarizing function to characterize the risk of a package with a single value.

   |input|function|output|
   |---|---|---|
   |**numeric vector** (0-1)|`summarize_risk`|**numeric** (0-1)|

## example package workflow using this structure

With all of these pieces in play, a common usage might look like this:

1. create a tibble column of package names

   |package|
   |---|
   |`"tools"`|
   |`"ggplot2"`|
   |...|
1. apply `as.pkg_reference` over column to create a column of `pkg_reference` objects

   |package|
   |---|
   |`<S3: pkg_install, pkg_reference>`|
   |`<S3: pkg_remote, pkg_reference>`|
   |...|
[...]

PR #28, opened 2019-07-09
added assessment function (see #27) for "has NEWS"

dgkf:
I implemented `get_*` style functions (renamed to `assess_*` as per #27) for determining whether a package has a NEWS file.

* [for remote package references](https://github.com/pharmaR/riskmetric/blob/6b7b75e63ca9ad8316eff161f617664eb00182be/R/pkg_metric_has_news.R#L15-L31), the CRAN webpage is parsed for a NEWS link 
* [for installed package references](https://github.com/pharmaR/riskmetric/blob/6b7b75e63ca9ad8316eff161f617664eb00182be/R/pkg_metric_has_news.R#L35-L42), the `system.file(package = <name>)` location is used as a search path for a `NEWS.*` file
* [for source package references](https://github.com/pharmaR/riskmetric/blob/6b7b75e63ca9ad8316eff161f617664eb00182be/R/pkg_metric_has_news.R#L46-L53), the source package directory is used as a search path for a `NEWS.*` file

I also added a [`score`](https://github.com/pharmaR/riskmetric/blob/6b7b75e63ca9ad8316eff161f617664eb00182be/R/pkg_metric_has_news.R#L95-L99) function dispatched on `pkg_metric_has_news` which returns `1` if any of the NEWS files are deemed valid (can be parsed by a respective `.Rd` or `.md` reader) or `0` otherwise.

There are a couple small utilities added for getting a list of non-installed CRAN package names and installed package names to make it easier to share examples (as shown below)

Some examples of how this might be used:

1. using a remote package:
   ```r
   pkg_name <- sample(riskmetric:::not_installed_packages(), 1)
   pkg <- as.pkg_ref(pkg_name)  
   assessment <- assess_has_news(pkg)
   assessment_risk_score <- score(assessment)
   
   cat(sprintf(
     "package '%s' has %s NEWS file", 
     pkg_name, 
     if (assessment_risk_score) "a" else "no"))
   ```

2. using an installed package:
   ```r
   pkg_name <- sample(riskmetric:::installed_packages(), 1)
   pkg <- as.pkg_ref(pkg_name)  
   assessment <- assess_has_news(pkg)
   assessment_risk_score <- score(assessment)
   
   cat(sprintf(
     "package '%s' has %s NEWS file", 
     pkg_name, 
     if (assessment_risk_score) "a" else "no"))
   ```

3. using a package source directory:
   ```r
   pkg_dir <- getwd()  # assuming working directory is riskmetric package directory
   pkg <- as.pkg_ref(pkg_dir)  
   assessment <- assess_has_news(pkg)
   assessment_risk_score <- score(assessment)
   
   cat(sprintf(
     "package source at '%s' has %s NEWS file", 
     pkg_dir, 
     if (assessment_risk_score) "a" else "no"))
   ```

For critique on the design and structure, please continue the discussion in #27, otherwise we can discuss implementation details in this PR


R/pkg_metric_has_news.R
```diff
@@ -0,0 +1,99 @@
...
+}
+
+
+
+#' @importFrom tools file_ext
+assess_news_in_dir <- function(x) {
+  news <- list()
+
+  # accommodate NEWS.Rd, NEWS.md, etc
+  news$files <- list.files(x, pattern = "^NEWS\\.*")
```
elong0527: It might be a feature or a bug in `list.files`. the output has been sort instead of matching the input package. You can see from below, `rtf` package (5th pacakge) have filename "NEWS", but it is in the first place in the output. 

```
> list.files(pkg$path[1:5], pattern = "^NEWS\\.*", full.names = TRUE)
[1] "/SFS/user/ctc/zhanyilo/R/covr/NEWS.md"    "/SFS/user/ctc/zhanyilo/R/gt/NEWS.md"     
[3] "/SFS/user/ctc/zhanyilo/R/janitor/NEWS.md" "/SFS/user/ctc/zhanyilo/R/rex/NEWS.md"    
[5] "/SFS/user/ctc/zhanyilo/R/rtf/NEWS"       
> list.files(pkg$path[1:5], pattern = "^NEWS\\.*")
[1] "NEWS"    "NEWS.md" "NEWS.md" "NEWS.md" "NEWS.md"
> pkg$path[1:5]
[1] "/SFS/user/ctc/zhanyilo/R/covr"    "/SFS/user/ctc/zhanyilo/R/gt"      "/SFS/user/ctc/zhanyilo/R/janitor"
[4] "/SFS/user/ctc/zhanyilo/R/rex"     "/SFS/user/ctc/zhanyilo/R/rtf"   
``` 

* in addition, is it possible to have `news` or `news.md` ? should we use `toupper` here?

R/pkg_metric_has_news.R
```diff
@@ -0,0 +1,99 @@
...
+        # NOTE: should we do validation of markdown format?
+        news$content[[i]] <- readLines(f)
+      }
+      news$valid[[i]] <- TRUE
+    }, error = function(e) {
+      news$valid[[i]] <- FALSE
+    })
+  }
+
+  # NOTE: should we test whether NEWS file is up-to-date with latest version?
```
elong0527: That would be great if we could handle it easily. Authors may have different way to write the version into the NEWS. (or we identify a pattern like `[0-9].[0-9].[0-9]` to match with the DESCRIPTION part?) 

As long as they have a NEWS file, I would consider the author is responsive.

R/pkg_metric_has_news.R
dgkf: My intention here was that the `assess_*` functions would be called only with a `pkg_reference` object as input, such that `x` would be a single package directory, such that `list.files(...)` would give all available NEWS files from the package. 

For the time being, I think NEWS must be capitalized (per the extensions manual, it also [must be a `NEWS.Rd` file](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#index-news), but I believe CRAN accepts a `NEWS.md` file these days it it seems to be the growing preference). Given how flexible the format is here, I wouldn't be surprised if a lowercase version was possible in the future, but I'd prefer not to try to address it unless that becomes a community standard.

R/pkg_metric_has_news.R
dgkf: Would it be okay to leave this for a future update? I'd prefer not to go down the rabbit hole of trying to validate all the different formats of NEWS files (`NEWS.md`, `NEWS.Rd`, `NEWS.html` (for remotely accessed NEWS files)) quite yet. 

Even those various text formats don't have any consistent syntax that is enforced across all packages and I'd be afraid that we could potentially invalidate otherwise good packages by being too prescriptive about what constitutes being "up-to-date".

elong0527: Yeah, I agree we can leave this for future. And we should keep the program neat and simple.

R/pkg_metric_has_news.R
elong0527: I see. Could we add a checking to ensure the length of `x` is 1 if it is the intention.

dgkf: I'd prefer not to add a check around it for now. Given that this is being used in a non-exported function and is only called internally using paths extracted from a `pkg_ref` object, I don't see it as a big threat. If we ever think about vectorizing `assess_*` functions, there would be a number of places that would have to be modified upstream of this function so it would be unavoidable to do some modifications anyway.

elong0527: @dgkf , I agree. Thanks for the clarification. I think we can complete the review after passing the Travis CI check.

PR #30, opened 2019-08-04
Assess news current

dgkf:
* added `assess_news_current`to assess whether NEWS includes update for recent version
* added `assess_export_help` to assess whether package has documentation for exported objects
* added a ton of documentation for our own functions
* commented out a lot of the source downloading code that @kkmann had added to get our tests passing and to get the package stable before R/Pharma. The structure of the package has changed slightly since that code was added and it's not causing minor issues.
* made some minor changes to vignettes to help pass R CMD Check


dgkf:
Hi @elong0527, did you have any chance to take a look at these changes? Would love to get your thoughts on how it's shaping up. If you don't have time to review, I'd prefer to move quickly with changes while we ramp up for R pharma so that we can do some testing prior to then.

elong0527:
Sorry for the delay. I was fully loaded for these days. Yeah, please go ahead and complete the necessary changes.

kieranjmartin:
Hi @dgkf as per your request I tried installing (I assumed you meant this branch as I imagine its most current). Some feedback:

There is a "riskmetric" vignette which is just calling the library. You should either delete or update

in quick-start the index at the top for score assess etc has the code version of the text rather than just the html link look, I think it should look the same as the other links.

"Finding a suitable reference for the package" I find a bit confusing to parse. Maybe "finding a good source for information on a package"

"class is an environment to allow for by-reference mutability so that metadata necessary to derive one metric can be reused by another without going out and fetching data all over again." this is a very technical explanation for a user; do they need to know that?

In general I think it would be better to say something like

"Risk_metric stores information on all packages on cran." (? don't know if true) "You can select which packages are of interest to you by providing pkg_ref the names of the packages you want. If there is a package which isn't contained in risk_metric you will need to provide a link to the package metadata, as in this example"

This is my guess of how the code works, as I haven't looked in detail!

(I didn't follow the full discussion on the creation of this environment, but I do wonder if at some point you're better just using R6 rather than creating your own lightweight version)

The rest of quick start looks good to me.



Extending risk metric has a typo:

"riskmetric is designed to be readily extensible. This is done through use of the S3 method dispatch system and a consciou acknowledgement "


This is all the time I have right now for review, but hope this is helpful, good luck!

dgkf:
Thanks @kieranjmartin - I really sincerely appreciate you carving out time to take a look as you're out the door on your holiday. I'll definitely take this feedback into account and try to polish the language in preparation of R/Pharma.

PR #68, opened 2019-11-25
add pkg_ref_cache.help_aliases.pkg_source fn

mpadge:
Hi @dgkf and others - great package! In my initial playing around, i noticed that the help assessment only worked for `pkg_install`, and just errored for `pkg_source`. This is an initial step to extract help aliases for the latter. It then just needs to be plugged in to a matching [`assess_export_help.pkg_source` function](https://github.com/pharmaR/riskmetric/blob/master/R/assess_export_help.R#L18) and should work, but I thought I'd leave this initial PR lightweight to start with.

### DIFF
--- a/R/pkg_ref_cache_help_aliases.R
+++ b/R/pkg_ref_cache_help_aliases.R
@@ -22,8 +22,9 @@ pkg_ref_cache.help_aliases.pkg_source <- function (x, name, ...) {
                                                 rd [grep("^\\\\alias", rd)]))
                       man_name <- strsplit (strsplit (i, "\\/man\\/") [[1]] [2],
                                             "\\.Rd") [[1]]
-                      names (a) <- rep (man_name, length (a))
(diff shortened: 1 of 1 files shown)


mpadge:
Sweet, thanks! And yeah, you're right that it's pretty structurally complicated. How do you think it would be best for me to provide feedback? It certainly was a bit of a learning curve for a package that provides immediate output the utility of which appears to confusing until transformed by an additional few post-processing steps... Shall i open an issue and describe a bit there, and then we could work towards refining in further issues?

dgkf:
> @mpadge
> Shall i open an issue and describe a bit there, and then we could work towards refining in further issues?

That would be great. 

If you'd like, we'll also have our next sprint check-in on Dec 3rd (12-12:30 EST). That might be a good venue for discussion, and you'd be more than welcome to join.

PR #88, opened 2020-02-10
add _has_vignettes metric

elong0527:
@dgkf 

Could you help me understand the trick for the `assess` function? Here is the code I test, it is success in the `pkg_ref` but fail in `assess` part. 

```
library(devtools)
load_all()

tmp <- pkg_ref(c("EIX", "dplyr", "smcure"))
lapply(tmp, function(x) x$vignettes)
lapply(tmp, function(x) nrow(x$vignettes) )

assess_has_vignettes(tmp)
```


dgkf:
Looks good @elong0527 

The issue with `assess_has_vignettes` is that it doesn't know what to do with a `list_of_pkg_ref` objects, which `pkg_ref` returns when provided with a vector of package names. This might speak to the need to be more type-stable with our core constructors to avoid this type of confusion.

Both of these should work:

```r
# applying `assess_has_vignettes` to each `pkg_ref` one-by-one
tmp <- pkg_ref(c("EIX", "dplyr", "smcure"))
lapply(tmp, assess_has_vignettes)
```

or

```r
# use the `assess` helper to apply all `assess_*` functions and return an assessment data frame
assess(pkg_ref(c("EIX", "dplyr", "smcure")))
```

I do encounter an error when running `assess_has_vignettes(pkg_ref("smcure"))` (I don't have this package installed so it gets referenced as a `pkg_cran_remote` class). 

I have some design and stylistic comments that I'll leave in my review, but I think it's already quite nice where it is.

R/pkg_ref_cache_vignettes.R
```diff
@@ -0,0 +1,72 @@
...
+#' @param path a package directory path expected to contain Vignettes files
+#'
+#' @return a data frame of Vignettes files
+#'
+vignettes_from_url <- function(url) {
+
+  file_path <- xml2::read_html(url) %>%
+    rvest::html_nodes("a") %>%
+    rvest::html_attr("href") %>%
+    grep("vignettes", x = ., value = TRUE)
```
dgkf: My preference would be to avoid using `dplyr` (and possibly `rvest`) internally as much as possible. These operations are pretty routine and should be quite manageable in base R. 

`xml2` provides some XPath based search that should be able to pull this information. It's not nearly as clean, but it would mitigate our dependencies. For an example of how to pull a link, check out the function `pkg_ref_cache.news_urls.pkg_cran_remote`. We're _almost_ to the point that we can cut out `dplyr` as a dependency, so I'd like to minimize relying on it where possible. Our only `rvest` code is some legacy stuff that's all been commented out at this point. 

If you'd prefer, I'd be fine pulling it in as is and filing another issue to cut down on the dependencies.

R/pkg_ref_cache_vignettes.R
```diff
@@ -0,0 +1,72 @@
...
+  folder <- c(source = "/vignettes", boundle = "/inst/doc", binary = "/doc")
+  files <- unlist(lapply(paste0(path, folder), list.files, full.names = TRUE))
+
+  if (!length(files)) return(data.frame())
+
+  file_path = unique(tools::file_path_sans_ext(files))
+  filename = basename(file_path)
+
+  res <- data.frame(filename = filename, path = file_path)
+  res <- res[tolower(res$filename) != tolower("index"), ]
```
dgkf: Instead of returning a `data.frame` which has rows of `filename` and `path`, perhaps this could be just as well represented as a named character vector where the values are the path and the names are the filenames?

I generally prefer to work with as minimal a data structure as possible which really helps to reduce complications if we end up needing to reference this in multiple places.

R/pkg_ref_cache_vignettes.R
```diff
@@ -0,0 +1,72 @@
...
+  file_path <- xml2::read_html(url) %>%
+    rvest::html_nodes("a") %>%
+    rvest::html_attr("href") %>%
+    grep("vignettes", x = ., value = TRUE)
+
+  filename <- tools::file_path_sans_ext(basename(file_path))
+  file_path <- paste0(url, "/", file_path)
+
+  res <- data.frame(filename = filename, path = file_path)
+  res <- res[tolower(res$filename) != tolower("index"), ]
```
dgkf: (see comment at line 46)

R/pkg_ref_cache_vignettes.R
```diff
@@ -0,0 +1,72 @@
...
+#'
+vignettes_from_url <- function(url) {
+
+  file_path <- xml2::read_html(url) %>%
+    rvest::html_nodes("a") %>%
+    rvest::html_attr("href") %>%
+    grep("vignettes", x = ., value = TRUE)
+
+  filename <- tools::file_path_sans_ext(basename(file_path))
+  file_path <- paste0(url, "/", file_path)
```
dgkf: I did get an error when using a `pkg_cran_remote` caused by this line, indicating that the lengths of `filename` and `file_path` were different. This happens when `file_path` is initially of length 0, meaning that `paste0(url, "/", c())` becomes just `{url}/`.

```suggestion
  file_path <- sprintf("%s/%s", url, file_path)
```

Reproducing the error:

```r
tmp <- pkg_ref("smcure")
class(tmp)
# [1] "pkg_cran_remote" "pkg_remote"      "pkg_ref"         "environment"  
assess_has_vignettes(tmp)
# Error in data.frame(filename = filename, path = file_path) : 
#  arguments imply differing number of rows: 0, 1 
```

R/pkg_ref_cache_vignettes.R
```diff
@@ -0,0 +1,72 @@
...
+#' Build a list of Vignettes files discovered within a given directory
+#'
+#' @param path a package directory path expected to contain Vignettes files
+#'
+#' @return a list of parsed Vignettes files
+#'
+vignettes_from_dir <- function(path) {
+  # accommodate unique vignettes files
+
+  folder <- c(source = "/vignettes", boundle = "/inst/doc", binary = "/doc")
```
dgkf: Maybe a typo?

```suggestion
  folder <- c(source = "/vignettes", bundle = "/inst/doc", binary = "/doc")
```

R/pkg_ref_cache_vignettes.R
elong0527: agreed and updated

R/pkg_ref_cache_vignettes.R
elong0527: agreed and updated

R/pkg_ref_cache_vignettes.R
elong0527: agreed and updated

elong0527:
after running `document()` function, I got an error message to rerun the ad-hoc code below. 

```
tmp <- pkg_ref("EIX", "dplyr", "smcure")
```

```
 Error in x %in% bioc_available$Package : 
  object 'bioc_available' not found 
```

while running `document()` function I also receive a warning message 
```
Loading riskmetric
Warning: [C:\Users\zhanyilo\docs\git\riskmetric\R\assess_downloads.r:5] @eval failed with error:
could not find function "assess_family_roxygen"
Warning: [C:\Users\zhanyilo\docs\git\riskmetric\R\assess_last_30_bugs_status.R:3] @eval failed with error:
could not find function "assess_family_roxygen"
```

Reference of all my ad hoc testing code.
```

library(devtools)
load_all()

pkg <- "survival"
path <- system.file(package = pkg)
vignettes_from_dir(path)

x <- pkg_ref("survival")
assess_has_vignettes(x)

x <- pkg_ref("EIX")
vignettes_from_url(x$web_url)

x <- pkg_ref("smcure")
assess_has_vignettes(x)

x <- pkg_ref("abtest")
assess_has_vignettes(x)



tmp <- pkg_ref("EIX", "dplyr", "smcure")
lapply(tmp, function(x) x$vignettes)
lapply(tmp, function(x) nrow(x$vignettes) )

assess_has_vignettes(tmp[[2]])
```

dgkf:
Ah, yeah- I got this too but forgot to mention it.

I renamed `assess_family_roxygen` to `roxygen_assess_family` (I think, but
not at a PC at the moment to check) to disambiguate it from the assess
functions.

There are two places where it has to be corrected if you want to go ahead
and fix them in the same pull request. If not I can fix it in a separate PR.

elong0527:
Thanks! roxygen_assess_family has been updated.

Any thoughts on the error message?

dgkf:
@elong0527 man, I somehow sneaked in a bunch of faulty code >.<. The `bioc_available` is related to a move to memoising the `bioc_available` data frame. This code is the culprit and has nothing to do with your pull request but rather an existing bug. 

https://github.com/pharmaR/riskmetric/blob/4d554ec82735cb47c87afd922284a47a68f75162/R/pkg_ref_class.R#L168-L169

In some un-merged code I have laying around I fixed it like this:

```r    
    } else if (x %in% memoise_bioc_available()[,"Package"]) {
      bp <- memoise_bioc_available()
      info <- bp[bp[,"Package"] == x,,drop = FALSE]
```

The rationale for the change was so that you don't go out and pull all the CRAN and bioconductor available packages until the first package requires that you do (which adds ~10s to each first run). Long term the most frequent use case will probably be when you have all available source code or at least a local library, so doing this for things like automated builds would unnecessarily lengthen run times.

elong0527:
updated. Shall we merge the branch into master?

PR #102, opened 2020-04-20
Issue 100 has bug reports url

### LINKED ISSUE #100: Metric for whether a BugReports URL is provided
We already have a `pkg_ref`field for `bug_reports_url`, but we don't wrap this in a `pkg_metric` or have a scoring mechanism. 

To complete this metric, we need to add
- `assess_has_bug_reports_url`
   - I don't think we need any `pkg_ref` dispatch for this as any sub-class-specific behavior is handled in `pkg_ref_cache.bug_reports_url`
   - **returns:** a `character` value containing the BugReports field contents wrapped in a  `c("pkg_metric_has_bug_reports_url", "pkg_metric", "character")` S3 class (constructed using `pkg_metric()`)
- `score.pkg_metric_has_bug_reports_url`
   - **returns:** a `logical` value indicating whether the package has a BugReports field filled in

`pkg_ref_cache.bug_reports_url` was already added as part of #48 

---

@elimillera volunteered to help with this one 👍 . Please let us know if you need any further clarification or if bandwidth to work on it is hard to come by.

elimillera:
This resolves #100.

I tested this on a couple of packages and it looks to be working as expected. Let me know if I am missing anything and I would be happy to change it!
I also updated the assess_test file to reflect the changes.

Also, ignore all the commits, RStudio did not like working with GitHub today.

### DIFF
--- a/R/assess_has_bug_reports_url.R
+++ b/R/assess_has_bug_reports_url.R
@@ -11,7 +11,7 @@ assess_has_bug_reports_url <- function(x, ...){
 
 # assign a friendly name for assess column
 attr(assess_has_bug_reports_url,"column_name") <- "has_bug_reports_url"
-attr(assess_has_bug_reports_url,"label") <- "presence of a bug_reports_url in repository"
+attr(assess_has_bug_reports_url,"label") <- "presence of a bug reports url in repository"
 
 #' Score a package for the presence of a bug report url
 #'
--- a/R/assess_has_bug_reports_url.R
+++ b/R/assess_has_bug_reports_url.R
@@ -1,0 +1,27 @@
+#' Assess a package for the presence of a url field where bugs can be reported.
+#'
+#' @eval roxygen_assess_family("has_bug_reports_url", 
+#' "a character value containing the BugReports field contents")
+#'
+#' @export
+assess_has_bug_reports_url <- function(x, ...){
+  has_bug_reports_url <- x$bug_reports_url
+  pkg_metric(has_bug_reports_url, class = "pkg_metric_has_bug_reports_url") 
+}
+
+# assign a friendly name for assess column
+attr(assess_has_bug_reports_url,"column_name") <- "has_bug_reports_url"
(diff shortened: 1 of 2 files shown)


R/assess_has_bug_reports_url.R
```diff
@@ -0,0 +1,27 @@
...
+#'
+#' @export
+assess_has_bug_reports_url <- function(x, ...){
+  has_bug_reports_url <- x$bug_reports_url
+  pkg_metric(has_bug_reports_url, class = "pkg_metric_has_bug_reports_url") 
+}
+
+# assign a friendly name for assess column
+attr(assess_has_bug_reports_url,"column_name") <- "has_bug_reports_url"
+attr(assess_has_bug_reports_url,"label") <- "presence of a bug_reports_url in repository"
```
dgkf, line 14: Very minor suggestion, but I think labels are better to be plain-english instead of referring to code objects

```suggestion
attr(assess_has_bug_reports_url,"label") <- "presence of a bug reports url in repository"
```

R/assess_has_bug_reports_url.R
```diff
@@ -0,0 +1,27 @@
...
+#'
+#' @eval roxygen_score_family("has_bug_reports_url")
+#' 
+#' @return A logical value indicating whether the package has a BugReports field
+#'   filled in
+#' 
+#' @export
+score.pkg_metric_has_bug_reports_url <- function(x, ...) {
+  # Return a score of TRUE if a bug report url is found, false otherwise
+  ifelse(!is.null(x), TRUE, FALSE)
```
dgkf, line 26: For `pkg_cran_remote` classes (I think also Bioconductor), `pkg$bug_reports_url` will be a zero-length character vector (`character(0L)`). This will return `TRUE` even if no valid `bug_reports_url` is found.

```suggestion
  length(x) > 0
```

elimillera:
Perfect, Thanks Doug! I'll also keep in mind bioconductor for future tests

PR #103, opened 2020-04-26
Adding code coverage metric

dgkf:
- Had to implement a 'custom' code method to test code coverage to avoid termination on error. It's practically identical to the base `covr::package_coverage` but skips the line that stops on errors.
- Added caching for both function- and expression-tallied coverage
- Added scoring for package coverage, returning total coverage as returned from `covr`
- Updated `summarize_risk` to handle `NA`s 
- Updated `.risk_weights` to include default weights for recently added metrics


dgkf:
Hmm... looks like I forgot to pull master before creating my branch... I'll rebase on master and create a new pull.

PR #107, opened 2020-05-08
use upper-case description filename; Fixes #106

### LINKED ISSUE #106: pkg_ref_cache_DESCRIPTION.R reads "description" instead of "DESCRIPTION"
I'm not sure if in some cases the DESCRIPTION file is lower-case, but I think these need to be swapped to upper-case. `pkg_ref("dplyr") %>% as_tibble() %>% assess()` results in the following warning: 
Warning message:
In read.dcf(file.path(x$path, "description")) :
  cannot open compressed file '~/my-R-libs/dplyr/description', probable reason 'No such file or directory'

dgkf:
- changed expected description file name from "description" to "DESCRIPTION"
- tested on macos (not-case sensitive file system) and linux (case-sensitive file system) to confirm changes work as expected on all platforms

### DIFF
--- a/R/pkg_ref_cache_DESCRIPTION.R
+++ b/R/pkg_ref_cache_DESCRIPTION.R
@@ -10,11 +10,11 @@ pkg_ref_cache.description <- function(x, name, ...) {
 
 
 pkg_ref_cache.description.pkg_install <- function(x, name, ...) {
-  read.dcf(file.path(x$path, "description"))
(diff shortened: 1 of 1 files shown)


elong0527:
I have addressed it in pull request 101.

Could we merge 101 first or change the destination of this pull request to dev/website?

dgkf:
Ah, didn't see that - thanks @elong0527, I'll cancel this one.

PR #113, opened 2020-05-14
explicitly declare covr package coverage path; Fixes #103

dgkf:
- Adds explicit package path to `covr::package_coverage` call
- Added a small comment to explain why we pass this expression to the covr call for anyone that happens to stumble up on this code.


dgkf:
Thanks for the quick review, @elong0527 - going to go ahead and merge.

PR #114, opened 2020-05-22
fix pkg_ref_cache.name_first_letter

mpadge:
This fix is necessary because this line only matches dispatches on classes of `pkg_ref_cache` objects, so must at least include `.default`:
https://github.com/pharmaR/riskmetric/blob/b2c8f5097b6fc8c2b13a7ed6cc01dc22efc54bf2/R/pkg_ref_cache.R#L91
The general workflow will still work without this, but for example, calling the `print` method will ignore any new cached metrics.

---

Note that this is a draft PR because there's a bug that needs to be fixed first: the `bugs_status` assess returns an `<error>`, which causes `score()` to generate warnings.

### DIFF
--- a/vignettes/extending-riskmetric.Rmd
+++ b/vignettes/extending-riskmetric.Rmd
@@ -162,8 +162,10 @@ package <- pkg_ref("riskmetric")
 
 # hack in order to mutate package environment directly so nobody accidentally
 # publishes any personal info in their library path
+ip <- data.frame(installed.packages())
 invisible(riskmetric:::bare_env(package, {
-  package$path <- sprintf("/home/user/username/R/%s/Resources/library/riskmetric", rver)
+  package$path <- file.path (ip$LibPath[ip$Package == "riskmetric"], "riskmetric")
+  #package$path <- sprintf("/home/user/username/R/%s/Resources/library/riskmetric", rver)
 }))
 
 package
@@ -187,15 +189,32 @@ order of assessments without worry about how metadata is acquired.
 ## Writing a Metadata Cache
 
 Now, for our new metric, we want to cache the package name's first letter. We
-need to add a new `pkg_ref_cache` function for our field. Thankfully, any
-subclass of `pkg_ref` can access the first letter the same way, so we just need
-the one function.
+need to add a new `pkg_ref_cache` function for our field. These functions
+minimally require a `default` dispatch method like the following:
 
 ```{r}
 pkg_ref_cache.name_first_letter <- function(x, name, ...) {
-  substr(x$name, 0, 1)
+  UseMethod("pkg_ref_cache.name_first_letter")
+}
+pkg_ref_cache.name_first_letter.default <- function(x, name, ...) {
+  substr (x$name, 0, 1)
 }
 ```
+The `default` method can also be replaced by methods for particular types of
+`pkg_ref_cache` objects (such as `pkg_source` or `pkg_install`). This cached
+field will then appear in the object:
+```{r}
+package <- pkg_ref("riskmetric")
+package
+```
+We can see that the `name_first_letter` field appears, yet has not yet been
+evaluated. Once it has, the value will appear in full within the printed
+object:
+```{r}
+package$name_first_letter
+package
+```
+
 
 After adding this caching function, we need to make a small modification to
 `assess_name_first_letter.pkg_ref` in order use our newly cached value.


dgkf:
Good catch, @mpadge. Instead of making a `*.default` method mandatory, I'd prefer to accommodate non-S3-dispatched caching functions, and by extension a developer could choose to use S3 to cater the behavior based on class. 

The `available_pkg_ref_fields` are only used for the `.DollarNames` completions. I think the priority should be the developer experience and we can find a way to accommodate it for the object names and completions.

dgkf:
I took the opportunity to rework this a bit. The regex got a bit too convoluted for my liking when starting to add new rules, so I switched gears a bit and tackled it as a data processing problem.

I think this should allow for fields which don't need an S3 function. I already discovered we have a `source_control_url` that wasn't visible in `.DollarNames` because it's implemented without requiring class dispatch. 

```r
available_pkg_ref_fields <- function(x) {
  fs <- c(names(getNamespace(packageName())), utils::.S3methods("pkg_ref_cache"))

  # build data.frame by function name and S3 dispatch (to 2 levels of dispatch)
  fs_df <- as.data.frame(t(vapply(strsplit(fs, "\\."), `[`, character(3L), 1:3)))
  names(fs_df) <- c("func", "field", "class")
  fs_df <- fs_df[fs_df$func == "pkg_ref_cache",]
  
  fs_df <- Filter(function(method) {
    # filter for functions that implement an S3 for a subclass or don't have 
    # any dispatched functions (all dispatched classes are NA).
    any(method$class %in% c(class(x), "default")) || all(is.na(method$class))
  }, Filter(nrow, split(fs_df, fs_df$field)))
  
  fs <- names(fs_df)
  fs <- fs[order(fs)]
  fs
}
```

mpadge:
@dgkf That code still issues warnings:

``` r
library (riskmetric)
library (tibble)
library (magrittr)
p1 <- pkg_ref ("riskmetric")
a <- as_tibble (p1) %>%
    assess ()
a
#> # A tibble: 1 x 11
#>   package version pkg_ref             news_current has_vignettes bugs_status
#>   <chr>   <chr>   <list<pkg_ref>>     <list<pkg_m> <list<pkg_me> <list<pkg_>
#> 1 riskme… 0.1.0.… riskmetric<install> <lgl [0]>    0             <error>    
#> # … with 5 more variables: license <list<pkg_metric>>,
#> #   export_help <list<pkg_metric>>, downloads_1yr <list<pkg_metric>>,
#> #   has_news <list<pkg_metric>>, covr_coverage <list<pkg_metric>>
a$covr_coverage
#> <list_of<pkg_metric>[1]>
#> [[1]]
#> <pkg_metric_error in UseMethod("assess_covr_coverage"): no applicable method for 'assess_covr_coverage' applied to an object of class "c('pkg_install', 'pkg_ref', 'environment')">
s <- score (a)
#> Warning in error_handler(x, ...): no available scoring algorithm for metric of
#> class "pkg_metric_error", returning default score of 0.

#> Warning in error_handler(x, ...): no available scoring algorithm for metric of
#> class "pkg_metric_error", returning default score of 0.

p2 <- pkg_ref ("<path>/<to>/<riskmetric_src>")
a <- as_tibble (p2) %>%
    assess ()
a$export_help
#> <list_of<pkg_metric>[1]>
#> [[1]]
#> <pkg_metric_error in UseMethod("assess_export_help"): no applicable method for 'assess_export_help' applied to an object of class "c('pkg_source', 'pkg_ref', 'environment')">
s <- score (a)
#> Warning in error_handler(x, ...): no available scoring algorithm for metric of
#> class "pkg_metric_error", returning default score of 0.

#> Warning in error_handler(x, ...): no available scoring algorithm for metric of
#> class "pkg_metric_error", returning default score of 0.
```

<sup>Created on 2020-05-25 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

dgkf:
These warnings are expected. Whether they're emitted to console can be handled through the `error_handler` arguments to `asses` and `score`, accepting a function from the  `assessment_error_*` and `score_error_*` families respectively. You can silence, raise as warnings or trigger errors by using different handlers.

These stem from our incomplete coverage of all metrics across all `pkg_ref` subclasses. The `covr_coverage`, for example, we only implement for `pkg_source` references. And `export_help` is only implemented for `pkg_install` references. When `riskmetric` attempts to run an assessment that hasn't been implemented it triggers these warnings.

These warnings seem to be causing a lot of confusion, however. Initially the thinking was that we wanted to be abundantly transparent when things might be behaving unexpectedly, but I don't think we're effectively hitting that mark. I'm entirely open to suggestions for improvement.

mpadge:
Okay, that makes sense then, thanks. I think the confusion arose on my part from the fact that all `pkg_ref.pkg_install` objects will lack `covr_coverage` elements, and these will always generate warnings. As a user, I would want all default options to work without such warnings, which would either mean that the `pkg_ref.pkg_install.covr_coverage` entry be something other than an error, or that the warning from `assess` of such not be a warning (or both). As this is in fact entirely standard and accepted behaviour, I would suggest both would be appropriate. Something like `pkg_ref.pkg_install.covr_coverage` simply being `NA`, and `assess` then including some kind of default statement identifying that the field is empty. Same comments apply to second example, where `assess.pkg_source.export_help` generates a warning, because `pkg_ref.pkg_source.export_help` generates an error. This behaviour would always arise, and so ought never really be either errors or warnings, i would suggest.

dgkf:
`covr_coverage` is a tricky one and will force us to rethink a bit of the architecture. It depends on having the source code (including tests) as well as an installed version of the package. This breaks our model of either having the source code _or_ an installed version. 

We chatted in our sprint today about what the process might look like. I think we'll eventually need a way for a `pkg_source` to go through `R CMD BUILD && R CMD INSTALL` in order to associate a temporary install with the `pkg_source` object. Then we could run the coverage in a separate R process, loading the newly installed package from a temporary library before running building its coverage.

dgkf:
> `pkg_ref.pkg_install.covr_coverage` simply being `NA`

So far the working model has been to not implement fields which aren't applicable. From a developer perspective I prefer the simplicity this way. 

Instead, I think I'd prefer to introduce a `pkg_metric_na` subclass (similar to `pkg_metric_error` subclass that is used to propagate errors). This way we can implement assessments as explicitly not applicable as opposed to treating them all as not-yet-implemented and throwing a warning.

mpadge:
Great idea @dgkf - a `pkg_metric_na` subclass would be a fine solution indeed. As for tests on installed packages: That is something we are going to be tackling, and is relatively straightforward but needs one big change to your current approach in the form of locally caching downloaded objects (in this case, `.tar.gz` bundles from CRAN). That changes everything from your side, because that is arguably proper "caching", as opposed to current usage of that term through package documentation to refer to lazy evaluation of S3 classes. There is nothing wrong with your current terminology, but the minute you also implement actual caching (local storage of otherwise remote objects), then you need to disambiguate and that ain't so straightforward. Anyway, those are all issues i intend to open here in due time. In the meantime, tests for `pkg_install` classes are:

(1) Very easy once you've got local caching because you can just extract the `./tests` directory and run it on the installed package (after dealing appropriately with tests calling non-exported functions, which is all do-able); and
(2) Extremely important because absence of tests is one of the most reliable red flags of risk. In my opinion, any assessment of risk absolutely has to consider tests, and in particular tests for the host of less recent and more established CRAN packages which will all be `pkg_install` objects and yet simply do not have any tests at all.

In the meantime, you should feel free to close this PR if you want, and I'll go back to the proverbial drawing board and revise what should actually be PR'ed here.

dgkf:
I'm going to go ahead and close this. I incorporated the changes to `available_pkg_ref_fields` in a separate PR. If there are other additions we need to make from this PR, please feel free to reopen or open a new PR.

PR #115, opened 2020-05-23
Dependency dplyr tibble magrittr janitor

elong0527:
remove dependency of 

dplyr 
tibble 
magrittr
janitor


R/utils_memoised.R
```diff
@@ -35,7 +29,7 @@ memoise_cran_mirrors <- memoise::memoise({
   # NOTE: might need to implement actual caching to avoid inconsistent behavior
   # when run with spotty internet
   function(all = TRUE, ...,
-    host_exists = !is.null(curl::nslookup("cran.r-project.org", error = FALSE))) {
+           host_exists = !is.null(curl::nslookup("cran.r-project.org", error = FALSE))) {
```
dgkf, line 32: Not my preferred style, but until (if?) we impose a style with lintr I think we can accept anything sensible. 😛

R/utils_memoised.R
```diff
@@ -1,24 +1,18 @@
...
-        janitor::clean_names() %>%
-        janitor::remove_empty_cols()
-    }
+    cran <- cran[, - which(names(cran) == "MD5sum")[1]]
+
+    cran_name <- tolower(names(cran))
+    cran_name <- gsub("[-/@ ]", "_", cran_name)
+    cran_name <- gsub(".", "_", cran_name, fixed = TRUE)
+
+    names(cran) <- cran_name
```
dgkf, line 12: Instead of doing all of this preprocessing, I'd be totally fine having "messy" column names and the "MD5sum" column. I appreciate you going through and reproducing the functionality, but I think it might be better just to return the result of `tools::CRAN_package_db()` directly. 

This just reduces any separation from the raw data so that if changes get introduced to the result of the CRAN db we don't have to trace the issue through our preprocessing. 

```suggestion
```

I think there are only two lines that would be affected by this change.

#### `R/utils.R` (L#36):

```diff
- cran_packages <- crandb[["package"]]
+ cran_packages <- crandb[["Package"]]
```

#### `R/cran.R` (L#12):

```diff
- with(memoise_cran_db(), version[match(pkgname, package)])
+ with(memoise_cran_db(), version[match(pkgname, Package)])
```

R/utils_memoised.R
elong0527: agree, we can have an issue open to go through lintr for a global check.

elong0527:
@dgkf , could you merge the current changes to avoid potential conficts?

Issue #110, opened 2020-05-13
Implement a unit testing for additional packages checking


elong0527:
for unit testing: embed source code

elong0527:
@elimillera , somehow, I am not able to find you in the assignees. Would you be able to self-assign?

elimillera:
> @elimillera , somehow, I am not able to find you in the assignees. Would you be able to self-assign?

I must not have the right permissions to add myself. There should be a list of contributors in the repository settings under manage access. I think you or @dgkf should have access to that

elong0527:
yeah, I just send you a write access invitation.

PR #116, opened 2020-05-24
Adds two embeded packages and updates unit tests for assessments

elimillera:
I have created two synthetic packages that can mimic packages being tested from source. Next on the list is to create tests for pkg_ref_class and figure out HTTP mocking.

This is part of #110 but should not close it.


dgkf:
Updates look good to me. Got a few warnings and one skipped test, but I'm not so concerned with the test results. I'm happy to pull this in as a focal point to further unit test development and we can refine the tests themselves from there. 

Happy to accept the PR, but let's hold off on merging until we have the function name changes in place and make focus the necessary updates on open PRs to get the name change control.

Issue #108, opened 2020-05-12
`pkg_ref_cache.bug_reports_host` throws STRSXP error when $bug_reports_url is NULL


dgkf:
Encountered when running the following code:

```r
x <- pkg_ref("gsDesign") # pkg_install
x$bug_reports_url
#> NULL
assess_last_30_bugs_status(x)
#> Error in get_component_(x, 1) : Not compatible with STRSXP: [type=NULL]. 
```

I believe this is due to not handling `NULL` as a possible input in `urltools::domain`. 

Suggest changing from `x$bug_reports_url` returning `NULL` when no url is provided to instead return `NA_character_`.

PR #117, opened 2020-05-27
Rename high-level user-facing functions

dgkf:
- `score` => `pkg_score` for `tbl_df` and `list_of_pkg_metric`
- `score` => `metric_score` for `pkg_metric`
- `assess` => `pkg_assess`
- `summarize_risk` => `summarize_score`
- `pkg_score` applied to a `tbl_df` now also performs `summarize_score`, storing results in a hard-coded column, "pkg_score"
- `riskmetric.Rmd` and `extending-riskmetric.Rmd` updated to reflect changes
- errors triggered through `bug_reports`, first logged in #108 fixed
- related to #108, improved error messages for errors caused within `scrape_bug_reports`
- `score.pkg_metric_has_bug_reports_url` changed to  returning a numeric instead of a logical to be consistent with other scores
- added `%||%` (common shorthand for "if not NULL else") to utils
- added safeguard to `pkg_metric` constructor disallowing accidental `structure(NULL, ...)` which will throw a warning. We should still aim to avoid this, though it shouldn't crop up accidentally now.


elimillera:
It doesn't look like this was caused by this commit, I'm having the same issue in the master branch, but I'm having problems with `summarize_score`. I tried with an installed and remote pacakge.
> dplyr_ref <- pkg_ref("dplyr")
> dplyr_assess <- pkg_assess(dplyr_ref)
> dplyr_score <- pkg_score(dplyr_assess)
Warning message:
In error_handler(x, ...) :
  no available scoring algorithm for metric of class "pkg_metric_error", returning default score of 0.
> summarize_scores(dplyr_score)
Error in apply(data[names(weights)], 1L, `*`, weights) : 
  dim(X) must have a positive length

dgkf:
Thanks @elimillera - I'll see if I can reproduce this.

dgkf:
Okay - found the issue. `summarize_scores` was assuming the input was a data frame. Now that `pkg_score` operates on single package references (not in a data.frame), it fails on `summarize_scores` when it then returns a single package list of scores. I updated it so that it can dispatch on score structure - either a data.frame or list.

elimillera:
@dgkf Looks good! I've had some new ideas about how to test remote packages so I'm going to get working on that once all the changes are merged in

R/summarize_scores.R
```diff
@@ -44,6 +47,14 @@ summarize_risk <- function(data, weights = .risk_weights) {
   risk
 }
 
+#' @export
+summarize_scores.list <- function(data, weights = .risk_weights) {
+  weights <- weights[which(names(weights) %in% names(data))]
+  weights <- weights / sum(weights, na.rm = TRUE)
+
+  1 - sum(as.numeric(data[names(weights)]) * weights, na.rm = TRUE)
```
elimillera, line 55: Sorry one last thing, I think we should return the score here and not the inverse of it. Now that we have renamed from summarize risk to summarize score, my intuition is that a higher scoring package would be better.

dgkf: Good catch, I was thinking the same thing. I'll opt not to invert it now that we're not calculating "risk".

Issue #124, opened 2020-06-03
Include weight for downloads_1yr in .risk_weights 


ndelcamp:
The downloads_1yr metric is not included in the risk calculation because it is not included in `.risk_weights`.

dgkf:
Another good find. This vector often gets overlooked as new metrics get added. 

I think we need to consider a better mechanism of setting defaults than building the vector manually. Perhaps a function which defaults to a weights of `1` for every assessment, but which accepts named parameters to overwrite this default would be more flexible and maintainable.

marlycormar:
Hi @dgkf and @elong0527, we should be able to close this issue now that PR #135 has been merged.

PR #135, opened 2020-07-08
Set default assessments weight to 1.

marlycormar:
Addresses issue #124


dgkf:
@marlycormar It just occurred to me based on an example from @jmanitz, that it might not be desirable to inject default weights. For example, if someone passes in a vector of weights, they likely don't want additional weights added in which would opaquely inject other metrics into their weighted score. 

Instead, perhaps we can apply these defaults only to the parameter defaults, such that if a numeric vector is provided by a user, they are not modified before calculating scores. 

I'm thinking we might aim for something like this:

```r
summarize_scores.data.frame <- function(data, weights = add_default_weights(data, .risk_weights)) {
  # ...
}
```

What do you think?

jmanitz:
thanks again for you help, I created a pull request with an example how to customize the weights

elong0527:
Suggestion:

1. remove default weights or set default weights as 0. 
2. When user provide weight, we would not want to inject additional weights. 
3.  create a `default_weight` (internal data or function) that allow user to modify and server. (ref

marlycormar:
Hi @dgkf, as we have been working on this PR since a few weeks back, I am adding an overview of the changes I made to aid you reviewing them.

- Created the following helper functions.
  - `check_weights`: Checks that the weights are numeric & non-negative.
  - `standardize_weights`: Calls `check_weights` first and then standardizes weights (0 to 1).

- Removed default weight values for missing assessments.

- Modified `pkg_score.list_of_pkg_metric` as suggested in today's call.

**Note**: As the user-provided weights' vector (if any) needs to pass through `standardize_weights`, I resorted to call `standardize_weights` in the body of `summarize_scores` rather than in its function arguments.

marlycormar:
Hi @dgkf, is there something missing in this PR? Happy to make updates.

R/summarize_scores.R
```diff
@@ -50,14 +49,12 @@ summarize_scores.data.frame <- function(data, weights = .risk_weights) {
 
 #' @export
 summarize_scores.list <- function(data, weights = .risk_weights) {
```
dgkf, line 51: I'd suggest adding the default value to add a default weight to our internal .risk_weights within the summarize_score function. This meets our needs for two different behaviors
1. When no weights are provided, fill in 1s for all missing columns, meaning we don't have to keep `.risk_weights` up-to-date as we add new metrics unless we want a weight other than 1.
2. Allows users to specify a partial set of weights, in which case missing columns will be unused (weight of 0)

```suggestion


add_default_weights <- function(data, weights) {
  cols_missing_weights <- setdiff(names(data), names(weights))
  weights[cols_missing_weights] <- 1
  weights 
}


#' @export
summarize_scores.list <- function(data, weights) {
  # when no weights are provided, add in a weight of 1 for all non-specified columns
  if (is.missing(weights)) weights <- add_default_weights(data, .risk_weights)
```

dgkf (changes requested):
This is looking good, @marlycormar 

I added in a suggestion to address Juliane's behavior recommendation. 

As well, I think we might want to modify `pkg_score` so that we don't pass columns "package", "version" or "pkg_ref" to `summarize_scores`. This will let us assume that _all_ of the columns in the dataset need to be scored.

```diff
pkg_score.tbl_df <- function(x, ..., error_handler = score_error_default) {
  assessment_columns <- get_assessment_columns(x)
  for (coln in which(assessment_columns)) {
    x[[coln]] <- vapply(x[[coln]],
      metric_score,
      numeric(1L),
      error_handler = error_handler)

    class(x[[coln]]) <- c("pkg_score", class(x[[coln]]))
  }

- x[["pkg_score"]] <- summarize_scores(x, ...)
+ ignore_cols <- c("package", "version", "pkg_ref") 
+ x[["pkg_score"]] <- summarize_scores(x[, !names(x) %in% ignore_cols], ...)

  # reorder columns so that metadata columns come first
  pkg_cols <- intersect(names(x), c("package", "version", "pkg_ref", "pkg_score"))
  x <- x[,c(pkg_cols, setdiff(names(x), pkg_cols))]

  x
}
```

R/summarize_scores.R
marlycormar: Hi @dgkf, thank you for reviewing the PR. I have a couple of questions.
1. Is `is.missing` a function in an R package? If not, we can set the weights arguments to `NULL` by default and check if it so in the body of the function:
   ```r
    summarize_scores.list <- function(data, weights = NULL) {
     # when no weights are provided, add in a weight of 1 for all non-specified columns
     if (is.null(weights)) weights <- add_default_weights(data, .risk_weights)
   ```

2. Should we also add these modifications to the `summarize_scores.data.frame` function?

3. Can we remove the `.risk_weights` list? If the user doesn't provide any weights, we could set the weight of each metric to 1. If the user does provide weights for some metrics, we could use those and ignore any missing metrics (weight = 0).

marlycormar:
> This is looking good, @marlycormar
> 
> I added in a suggestion to address Juliane's behavior recommendation.
> 
> As well, I think we might want to modify `pkg_score` so that we don't pass columns "package", "version" or "pkg_ref" to `summarize_scores`. This will let us assume that _all_ of the columns in the dataset need to be scored.
> 
> ```diff
> pkg_score.tbl_df <- function(x, ..., error_handler = score_error_default) {
>   assessment_columns <- get_assessment_columns(x)
>   for (coln in which(assessment_columns)) {
>     x[[coln]] <- vapply(x[[coln]],
>       metric_score,
>       numeric(1L),
>       error_handler = error_handler)
> 
>     class(x[[coln]]) <- c("pkg_score", class(x[[coln]]))
>   }
> 
> - x[["pkg_score"]] <- summarize_scores(x, ...)
> + ignore_cols <- c("package", "version", "pkg_ref") 
> + x[["pkg_score"]] <- summarize_scores(x[, !names(x) %in% ignore_cols], ...)
> 
>   # reorder columns so that metadata columns come first
>   pkg_cols <- intersect(names(x), c("package", "version", "pkg_ref", "pkg_score"))
>   x <- x[,c(pkg_cols, setdiff(names(x), pkg_cols))]
> 
>   x
> }
> ```

Hi @dgkf, this is a great idea. I just made this change to the code.

R/summarize_scores.R
dgkf: > 1. Is is.missing a function in an R package

My mistake - the base function is just `missing()`. It returns `TRUE` if no argument is passed to this parameter (even if a default is available) and `FALSE` otherwise. I always mix this one up.

> 2. Should we also add these modifications to the summarize_scores.data.frame function?

Yes, another oversight on my part. Thanks for catching that.

> 3. Can we remove the .risk_weights list?

Yeah, I think we could. The original idea was that we might want to provide a default weighting convention that might be more involved than just giving everything a weight of 1. We haven't really made major progress toward that, though, and I think it might be misleading users to think that we have some profound meaning behind our suggested weights when they're just placeholder. Let's get rid of it and we can reintroduce it if we establish a more meaningful set of suggested weights.

marlycormar: great, thanks @dgkf! I'll make those changes soon.

marlycormar:
Hi @dgkf, the PR is ready to be reviewed.

Similar to your suggested code change for the `pkg_score.tbl_df` function, I also removed the columns `c("package", "version", "pkg_ref", "pkg_score")` from the `data` input of `summarize_scores` (this is important for when `summarize_scores` is called directly instead of through `pkg_score`). We can improve both code snippets if we could assume that the columns of interest are only those of class `pkg_score`. Is this a valid assumption?

dgkf (approved):
Looks good, @marlycormar! Thanks for sticking with this one. 

I think we learned a lot along the way and had some new use cases pop up in the middle of development. I really appreciate you rolling with the changes and helping to provide a focal point for us to hash out exactly what we want from our user experience.

PR #137, opened 2020-07-22
Dev/issue 131

emilliman5:
Addition of R CMD checks (caching and assessment) for  cran_remotes, bioc_remotes and source. It does not appear possible to run R CMD check on an installed package so I set a default pkg_ref method to return `NA`. 

I am wondering if `R CMD check` should be a separate assessment from CRAN/Bioc check summaries. 
  1. `R CMD check` can be run on any pkg_ref subclass (except pkg_install), in theory. 
  1. `R CMD check` and the CRAN/Bioc check summaries provide different information, IMHO
        1. `R CMD check` checks  whether the package will install on your system and if there are any potential problems with the install. For example, when I run `R CMD check` I get a warning for every package because I do not have the ‘qpdf’ package for checks on size reduction of PDFs.
        1. CRAN/Bioc checks summaries provides a snapshot of how a package would install on numerous systems, i.e. its portability. This to me is a different, albeit important potential metric. To me a package that is portable is of lower risk than one that is less portable.
  1. I thought about pulling individual checks from CRAN to match with the `R CMD check` but decided against it since we would have to match the users environment with the environments that CRAN runs the checks on and assume that they are the same which seems counter to proper risk assessment principles.


dgkf:
Thanks @emilliman5 - this looks really promising. Just some notes from a cursory look at this work.

### Digging into the `check` result on a `pkg_source`

First - there's a bunch of data bundled in the return of `devtools::check`. It prints out prettily to console, but if you probe the return object it's got a lot of more granular data tucked away inside.

```
> x <- devtools::check("../riskmetric")
> attributes(x)
# $names
#  [1] "stdout"      "stderr"      "status"      "duration"    "timeout"     "rversion"    "platform"    "errors"      "warnings"   
# [10] "notes"       "description" "package"     "version"     "cran"        "bioc"        "checkdir"    "test_fail"   "install_out"
```

We could tease apart many of these as separate `pkg_ref_cache` fields. They could simply deconstruct pieces of this object. I've never played around with these return data types before, so this is all new to me and I'm happy to find they seem to be very accessible for further exploration.

### Structuring `pkg_remote` results

It seems we can get some of this info from CRAN, but we'd have to consider which "flavor" (OS) we'd want to pull info from. One design philosophy that we try to adhere to is structuring `pkg_ref_cache` results identically across class-specific methods. You could consider them like a common ground - a way of converging on a data structure independent of the source of the information. 

You already pull scrape some data from CRAN/Bioconductor, but we should ensure that they are structured in a similar way to the result of `pkg_ref_cache.cran_checks.pkg_source`. Right now it seems that the `pkg_cran_remote` and `pkg_bioc_remote` objects are structured as data frames while the `pkg_source` version is a list. Most likely, I think we should deconstruct the result of the `pkg_source` into some of its fields and only try to match a subset of these through CRAN/Bioconductor - most likely just the `stdout`, `errors`, `warnings` and `notes` that are readily available on the respective web pages. 

### On `pkg_install`

Like you said on our sprint call, I can't seem to find any way to recover the checks from an installed package. The best I can think of would be to try to pull the checks from an identical version on CRAN, though at that point I'm not sure I'd make that a priority. Although unlikely, there are no guarantees that a package with a particular version number installed locally was actually the same as the version we might scrape from CRAN. 

### Console log

Currently accessing the `$cran_checks` field will print out to console. So far we've tried to keep everything silent by capturing or redirecting output. However, I've been reconsidering this decision recently. In conjunction with the shiny app, I think it would be preferred to get the console output as all the metrics are being run. One complication is that we gather a lot of the data via `pkg_ref_cache` methods, not the metrics themselves, so we need to think about how best to collect these and either sink them to console or a reactive shiny element.

emilliman5:
@dgkf Thanks for the input. 

I will work on deconstructing the R CMD check output into various cache fields for future assessments. 

I will work more on unifying the **pkg_remote** checks with the **pkg_source** checks, either by trying to scrape the detailed check reports on cran or tallying errors/warnings/notes

What do you think about creating a "remote_checks" result (from CRAN/Bioc/etc) and a local "R CMD check" result? 
This would make the results easier unify.
And allow us some extra information. CRAN/Bioc checks as implemented assess portability or provide initial/cursory info regarding R CMD check. While R CMD check would check against your system before install, and therefore a bit more rigorous. If we download the source during caching we could run `R CMD check` on **pkg_remote** (this was a use case you had mentioned).

Lastly, if we were to pull detailed checks from CRAN Mark P mentioned the `cchecks` database to access archived checks from previous package versions. Would this be something we want to incorporate? CRAN seems to only keep the most recent checks, which means we will most likely be mismatched on package version if we only rely on CRAN.

elong0527:
Suggestions:

1. a flow chart to visualize the structure.
2. make type consistent in `pkg_ref`.

emilliman5:
Hi everyone,

Here is my breakdown of the `R CMD check` output, I've tried to organize things based on **what** was checked so that each line item from the output ends up in one category. But there are also opportunities to categorize to the **type** of check (e.g. dependency checks)

1. **Declarations:**
   + using R Under development (unstable) (2020-08-08 r78985)
   + using platform: x86_64-pc-linux-gnu (64-bit)
   + using session charset: ISO8859-15
   + this is package 'ggplot2' version '3.3.2'
   + package encoding: UTF-8
1. **Data:**
   - checking contents of 'data' directory
   - checking data for non-ASCII characters
   - checking data for ASCII and uncompressed saves
1. **Vignettes:**
   - checking files in 'vignettes'
   - checking for unstated dependencies in vignettes
   - checking package vignettes in 'inst/doc'
   - checking re-building of vignette outputs
   - checking PDF version of manual
1. **Rds:**
   - checking Rd files
   - checking Rd metadata
   - checking Rd line widths
   - checking Rd cross-references
   - checking for missing documentation entries
   - checking for code/documentation mismatches
   - checking Rd \usage sections
   - checking Rd contents
1. **Examples:**
   - checking examples
   - checking for unstated dependencies in examples
1. **Function calls:**
   - checking use of S3 registration
   - checking S3 generic/method consistency
   - checking replacement functions
   - checking foreign function calls
1. **Code errors:**
   - checking R files for non-ASCII characters
   - checking R files for syntax errors
   - checking R code for possible problems
1. **Unit tests:**
   - checking for unstated dependencies in 'tests'
   - checking tests
1. **Package structure:**
   - checking for file 'ggplot2/DESCRIPTION' ... OK
   - checking if this is a source package ... OK
   - checking if there is a namespace ... OK
   - checking for executable files ... OK
   - checking for hidden files and directories ... OK
   - checking for portable file names ... OK
   - checking for sufficient/correct file permissions ... OK
   - checking serialization versions ... OK
   - checking package directory ... OK
   - checking for future file timestamps ... OK
   - checking 'build' directory ... OK
   - checking DESCRIPTION meta-information ... OK
   - checking top-level files ... OK
   - checking for left-over files ... OK
   - checking index information ... OK
   - checking package subdirectories ... OK
   - checking installed files from 'inst/doc' ... OK
   - checking for non-standard things in the check directory
1. **Package dependencies:**
   - checking dependencies in R code
   - checking package namespace information
   - checking package dependencies
1. **Install/load/unload:**
   - checking whether package 'ggplot2' can be installed ... OK
   - checking whether the package can be loaded
   - checking whether the package can be loaded w/ stated dependencies
   - checking whether the package can be unloaded cleanly
   - checking whether the namespace can be loaded w/ stated dependencies
   - checking whether the namespace can be unloaded cleanly
   - checking loading without being on the library search path

####Question 1: How do we organize this data within the `pkg_ref`?
If we were to create a single "RCMDcheck" slot with these (or some variation) sub categories embedded within, we then an assessment of the `R CMD check` would be very intuitive, but might obscure the information for other assessments. 

We could also create a pkg_ref slot for each of these categories (or variations thereof) which would keep the pkg_ref data a bit flatter, which may make it easier to use an individual slot for more than one assessment.

####Question 2: How do we want to use this information for assesments? (Maybe this is should be first question)
Do we want a single assessment for the totality of running `R CMD check` or do we want an assessment for each of these groups (or variations)? Having both would seem redundant, but could allow for a "simple" vs "detailed" assessment mode...

emilliman5:
This comment is a work in progress...

My thoughts on R CMD check flow for package development. 
1. Write code
1. Run R CMD check locally intermittently to catch problems with package
1. Finish writing code, tests, tests and documentation
1. Run R CMD check one last time to make sure there are no problems.
1. If no problems then submit to CRAN/Bioc
1. Remote repository runs R CMD check on their various R environment flavors. (CRAN has 12 flavors, Bioc has 3)
    1. CRAN provides a summary of the checks [like so](https://cran.r-project.org/web/checks/check_results_DirichletReg.html)
        1.  Screen Shot 2020-08-24 at 1 46 36 PM
        1. Each Status entry links to the detailed R CMD check output you are used to seeing when running it locally.
        1. The status is the marked according to the following order Error > Warning > Note > OK
1. If they all pass then package is accepted and published.

### Idea 1:
1.  Create a `pkg_ref` slot for the cran/bioc check summary (implemented in current PR) with the tally of errors, warnings, notes and Ok's for all flavors, in the above example the pkg_ref slot data would look like `c("OK" = 12)`
    1. We could and add a "local flavor" entry with a summary of the status has Error > has Warning > has Note > Ok and add to the tally above.
1. Create a slot for `pkg_ref.source` (and possibly `pkg_ref.remote`) to store the output of `R CMD check`: specifically any reported errors, warnings and notes. 
    1. An example for the above package source: `list(errors = character(0), warnings = "checking compiled code ... OK\n WARNING\n‘qpdf’ is needed for checks on size reduction of PDFs", 
    notes = character(0))`
    1. For `pkg_ref.installed` we could potentially use `sessionInfo()` to match to a CRAN/Bioc flavor and pull the detailed `R CMD check` output from CRAN.
    1. CRAN only maintains the output for the most recent check (i.e. the most recent package version and most recent R version) which may not match the users package version or the user's R version.

### Idea 2:
1. Create a `pkg_ref` slot to hold named list of all of the detailed check outputs (or just the reported errors, warnings and notes or an Ok see above example Idea 1 2.1).
     1. We could then add a "local" entry to above object with the same summary of errors, warnings, notes or Ok.
          1. For example: `list(errors = character(0), warnings = "checking compiled code ... OK\n WARNING\n‘qpdf’ is needed for checks on size reduction of PDFs", 
    notes = character(0))`
    1. This structure would allow us to tally status across platforms.
    1. For installed packages we could attempt to match the user's environment with a CRAN flavor
    1. We will have to parse the detailed CRAN/Bioc checks ourselves (not necessarily hard or easy but will be some work.

emilliman5:
this PR has been super-ceded by #170 and #169

PR #153, opened 2020-09-08
Resolves #108, fixes reference to missing field in vignettes_from_html

### LINKED ISSUE #108: `pkg_ref_cache.bug_reports_host` throws STRSXP error when $bug_reports_url is NULL

dgkf:
@emilliman5 - I reviewed the bug report in #108. Although I think I had previously fixed the STRSXP error that was being thrown as part of #117, I think that other bugs had been introduced causing this particular case to continue to throw misleading errors.

If you have the bandwidth, I would appreciate if you could rerun your example from #108 and confirm that it's working now. For this particular package, `gsDesign`, `pkg_assess` should report that the DESCRIPTION does not include a BugReports field.


emilliman5 (approved):
Hi @dgkf, I ran the example in #108 and get desired behavior (see below)

```
x <- pkg_ref("gsDesign") # pkg_install
x$bug_reports_url
#character(0)
assess_last_30_bugs_status(x)
# Error in scrape_bug_reports.default(x, ...) : 
#  package DESCRIPTION does not have a BugReports field 
```

PR #157, opened 2020-09-14
Adding webmocking for web requests

dgkf:
## Overview

Since a LOT of files got changed, I want to direct you to the most important parts:
- `setup_mock_web_requests.R` has the code for setting up the various http endpoints and mocked responses
- `setup_test_packages.R` is where I create some packages for use throughout testing. This is sort of a merge of code I had previously and the prior `setup-xyz.R`
- A bunch of local files saved as prepared mock reponses are stored in `tests/testthat/test_webmocks/`
- test packages have been moved to a directory in `tests/testthat/test_packages/` and have been renamed to reflect a -"good" and -"bad" naming scheme used throughout for all `pkg_ref` subclasses for testing.
- snapshots have not been updated

If you can run the tests and things don't fail catastrophically, I think we're in good shape. If anything goes wrong, you should see errors suggesting that a stub web mock hasn't been set up for a given http request, but hopefully I've caught all those.

### metric updates
- A few internal functions have changed from using `curl` to `httr` such that we can intercept web requests using `webmockr` and avoid hitting public APIs during testing runs
- Created a set of stub `pkg_refs` for testing, characterized as "good" or "bad" to try to test success and failure modes of each metric
- A minor update to `assess_has_vignettes` which would throw errors for remote `pkg_ref`s

### testing updates
- Added stubs for various cran- and github-related web requests that we issue throughout testing
- Added local versions of the http response for CRAN and GitHub responses
- Added function to generate cranlogs downloads response
- *TODO:* split cran responses into a representative "good" and "bad" remote package
- *TODO:* add bioconductor package testing

### tests
- Rewrote most tests to use these new, webmock'ed packages
- *TODO:* Update snapshots for snapshot tests after updates

---

Unit tests _almost_ pass, with only the snapshot tests now failing. I wanted to get input on the direction before making changes to the snapshots.


dgkf:
@elimillera - I know you're busy at the moment, so I'm going to go ahead and merge this in to continue working on getting the testing working. We can do a full review between all contributors to the feature branch once we pull into master, but I don't think we need to bog ourselves down in incremental reviews until then.

Issue #92, opened 2020-02-21
BugReports error for pkg_install class object


dgkf:
I stumbled upon this specifically with the `nlme` package, which I have installed in my local library.

```r
x <- pkg_ref("nlme")
stopifnot(inherits(x, "pkg_install"))  # check to confirm it's installed
x$bug_reports
```

Interestingly, `x$bug_reports_host` is somehow just set to `bugs`.

emilliman5:
This behavior is because bug reports for nlme (and other official R packages) are hosted on a subdomain of the r-project.org site, i.e. https://bugs.r-project.org. Instead of `extract_host()` we could use `domain()`

elong0527:
Agree to use  `domain` 

add testing cases to ensure the change works for `github` and `gitlab`

PR #158, opened 2020-09-16
modified the paring of the bug url to grab the last domain before the…

emilliman5:
… TLD, this should be more robust to bug urls that use a subdomain instead of domain+path. I opted to keep the same intended behavior because the bugs_url_host is used for S3 dispatch of `scrape_bug_reports`. This is in reference to #92

### DIFF
--- a/R/pkg_ref_cache_bug_reports_host.R
+++ b/R/pkg_ref_cache_bug_reports_host.R
@@ -12,5 +12,5 @@ pkg_ref_cache.bug_reports_host <- function(x, ...) {
 #' @importFrom urltools host_extract domain
 pkg_ref_cache.bug_reports_host.default <- function(x, ...) {
   if (is.null(x$bug_reports_url)) return(NULL)
-  host_extract(domain(x$bug_reports_url))$host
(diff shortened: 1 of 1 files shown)


R/pkg_ref_cache_bug_reports_host.R
```diff
@@ -12,5 +12,5 @@ pkg_ref_cache.bug_reports_host <- function(x, ...) {
 #' @importFrom urltools host_extract domain
```
dgkf, line 12: Let's remove the `host_extract` function from our namespace now that we're no longer using it while we're making these edits.

```suggestion
#' @importFrom urltools domain
```

dgkf (changes requested):
Thanks @emilliman5! Looks good. Let's pull it in with just the mentioned suggested change.

As a follow up, I think that we might we might want to transition to using the full domain for dispatch. This would help to future-proof us against a possible case where bugs might be captured in multiple subdomains. In that case, we might also need to preprocess the domain name (perhaps replacing all non-alphanumeric characters with`_`'s) to avoid issues with dispatch, but let's hold off on that change for a future issue.

dgkf (approved):
Looks good, I'll pull it in.

PR #159, opened 2020-09-17
various houskeeping items to clean up tests

dgkf:
Mostly organizational changes in here.

### Snapshot issues encountered:
1. Order of `pkg_ref` fields is critical to getting passing diffs. In a couple runs I encountered outputs that reordered the fields and diffs would get issued for just rearrangement, despite having no structural difference to the objects. I can dig into how we can make the output more consistent, but it doesn't seem like an immediate priority unless we are sure snapshots are absolutely necessary.
2. html parsed using `xml2` is stored as a a memory address pointer in R which will inevitably change between snapshots. 
3. `pkg_ref`s contain paths which will be inconsistent across individual machines and will trigger snapshot diffs
4. message output isn't captured, meaning that `<pkg_ref>$covr_coverage` won't be diff'ed in any meaningful way
5. downloads are calculated from the `Sys.Date()`, and will vary depending on the day on which the test suite is run.

### Other notes:
I thought that the message output from `covr::package_coverage` might have been the source of some test issues, so I wrote a wrapper to capture the messages, warnings, outputs and errors during code evaluation. I think this will be helpful for eventually building an audit log of all the operations that were performed, but for now it's just used for suppressing messages. It turned out these messages were just being triggered when testthat tries to print out the contents of a `<pkg_ref>$covr_coverage` for snapshotting. Oh well, I think this function will be useful later.


dgkf:
@elimillera I've disabled snapshots. I outlined a number of potential pitfalls that we'll run into because of the various outside factors that impact the exact object we end up with when creating `pkg_ref`. After a bit more of a dive into this, I'm questioning how helpful this will be long term. 

If it's okay with you, I'd like to disable these for now until we have the builds passing and then we can revisit some of these considerations.

elimillera (approved):
These all look good. I can return to snapshots after we get the classic ones figured out. Definitely not worth the trouble right now. I'll start to review the webmocking and the test packages and start working through the assessments. Let me know if you were planning to do any that aren't github yet.

dgkf:
Thanks @elimillera - I'll pull it in and we can figure out what we need to do to work in snapshots after we got builds that pass.

PR #161, opened 2020-09-26
[WIP] Building out testing suite and transition to GitHub actions

dgkf:
- Migrate to GitHub actions
- Lay out framework for having mock packages embedded in testing suite for source package metrics
- Mock web requests for installed/remote package metrics
- Some reworking of web requests to use `httr` for all web requests allowing `webmockr` to intercept requests
- Minor bug fixes to address issues encountered along the way


dgkf:
Test fails should be resolved once [webmockr #106](https://github.com/ropensci/webmockr/issues/106) makes its way to CRAN. Alternatively, if we want things to work ASAP we can embed all the full uri's for each of the mocked urls, but my preference is toward keeping the webmocking fairly generic.

dgkf:
✅ green checkmark! ✅

elimillera (approved):
Changes look really good. It looks like some of the mocked requests get into the assessment results, but I can get started updated and writing more tests.

dgkf:
During our sprint planning meeting yesterday, we discussed how to approach this code review. Since it's quite large, we thought we could split up the review. Major changes are introduced in `./R/` and `./tests/`

### `./tests/` (@elimillera)
I see that you've already had the opportunity to comb through the changes, so I won't bother breaking them down. Thanks for the speedy review!

### `./R/` (@elong0527)
- Modified `scrape_bug_reports.github` so that it can handle BugReports urls in the form of `www.github.com/xxx/xxx/issues`, as well as just `www.github.com/xxx/xxx` (and extends to alternative url subpaths)
- Modified `pkg_ref_cache.covr_coverage.pkg_source` so that messages aren't printed to console from `covr::package_coverage` while running `pkg_assess`
- Modified functions which used `xml2` to use `httr` instead to make http requests because `httr` can be intercepted by `webmockr` for testing. Additionally, suppressed messages when `httr` has to assume a file encoding.
- Removed a lot of legacy code related to caching CRAN. These have long been unused.
- Updated `memoise_` functions to accept a local directory which can be used to load a local result - a poor man's "mocking" for things that we can't intercept with `webmockr`
- Added one large, but relatively unused function, `capture_expr_output`. It captures all console output (output, messages, warnings, errors, result value, and whether the result is returned invisibly) and handles replaying the output back to console. I'm hoping we can use this to build a log of all the assessment code we run if we need to generate something like an audit log. Right now it's only used for suppressing the output of `covr::package_coverage`, but we could also use it to capture the console output of cached package fields as they're evaluated.

R/pkg_metric_error.R
```diff
@@ -1,5 +1,7 @@
 #' A subclass wrapping an error with an additional parent class
 #'
+#' @param e an error object to capture
+#'
 pkg_metric_error <- function(e) {
```
elong0527, line 5: could we change the parameter from `e` to `error` ?

R/zzz.R
```diff
@@ -1,4 +1,7 @@
+#' @importFrom backports import
 .onLoad <- function(libname, pkgname) {
+  backports::import(pkgname, "isFALSE")
```
elong0527, line 3: I had limited knowledge for this line of code and need help to review this file.

R/utils_memoised.R
```diff
@@ -1,28 +1,24 @@
...
 #' @importFrom curl nslookup
+#' @importFrom memoise memoise
+#'
 memoise_cran_mirrors <- memoise::memoise({
   # add parameter such that memoised results rerun if internet availability changes
   # NOTE: might need to implement actual caching to avoid inconsistent behavior
   # when run with spotty internet
-  function(all = TRUE, ...,
-           host_exists = !is.null(curl::nslookup("cran.r-project.org", error = FALSE))) {
+  function(all = TRUE, ..., .local = getOption("riskmetric.tests")) {
```
elong0527, line 16: suggest to reverse the order of `...` and `.local`. 
suggest to avoid argument start with `.`

R/utils_memoised.R
```diff
@@ -1,28 +1,24 @@
...
+    if (!is.null(.local)) {
+      return(read.csv(
+        file.path(.local, "test_webmocks", "data", "cran_mirrors.csv"),
```
elong0527, line 19: `getOption("riskmetric.tests")` return `NULL` by default. 

We might need to consider if `riskmetric.tests` is clear enough to indicate a path. 

Does  `file.path(.local, "test_webmocks", "data", "cran_mirrors.csv")` always return correct path from user?

R/utils_memoised.R
```diff
@@ -68,18 +64,28 @@ memoise_bioc_mirrors <- memoise::memoise({
...
+        file.path(.local, "test_webmocks", "data", "cran_packages.csv"),
+        stringsAsFactors = FALSE)
+      db[, "Repository"] <- contrib.url(repos, getOption("pkgType"))
+      return(db)
+    } else if (is.null(repos)) {
       return(utils::available.packages(NULL))
-    else if ("@CRAN@" %in% repos)
-      repos[repos == "@CRAN@"] <- "https://cran.rstudio.com/"
+    } else if ("@CRAN@" %in% repos) {
+      repos[repos == "@CRAN@"] <- "https://cloud.r-project.org"
```
elong0527, line 86: what's the rational to change `cran.rstudio.com` and `cloud.r-project.org`?

Shall we add `repo` as an argument to allow user determine the source? For example, people might be interest in the available package from an RSPM snapshot.  https://packagemanager.rstudio.com/client/#/repos/1/overview

R/utils.R
```diff
@@ -84,3 +55,199 @@ with_unclassed_to <- function(x, .class = 1:length(class(x)), expr,
...
+#'   are issued
+#' @param env the environment in which \code{expr} should be evaluated,
+#'   defaulting to the calling environment.
+#' @param quoted whether \code{expr} is a quoted object and should be evaluated
+#'   as is, or whether the expression should be captured from the function call.
+#'   Defaults to \code{FALSE}, capturing the passed expression.
+#' @inheritParams base::sink
+#'
+#' @importFrom utils head tail
+#'
```
elong0527, line 76: Could you add an example for this function to help future developer use it in other functions?

R/utils.R
```diff
@@ -84,3 +55,199 @@ with_unclassed_to <- function(x, .class = 1:length(class(x)), expr,
...
+#' @param ... additional arguments unused
+#' @param sleep an \code{numeric} indicating a time to sleep between printing
+#'   each line to console. This can be helpful if the original output overwrites
+#'   valuable information in the log that is eventually overwritten and you
+#'   would like to watch it play out as it was formatted.
+#'
+#' @export
+#'
+#'
+print.expr_output <- function(x, cr = TRUE, ..., sleep = 0) {
```
elong0527, line 167: How will we need this function? I did not see it is used in other place. 

Shall the name be `print_expr_output`? Otherwise, would you plan to create a class called "expr_output" to use `print` method?

elong0527 (commented):
Thanks for initiating all testing cases. 

I have reviewed all code change in the `.R/` folder. It is in good shape.

* will need help for reviewing the change in `zzz.R` file. 
* need clarification on new functions in `util.R`.

R/zzz.R
dgkf: We use the function `isFALSE` at one point in the package. It's from `base`, but is only available from 3.4 onward. The `backports` provides functions provided by a later version of R for packages in older version of R. 

We could reimplement it, but `backports` is quite a lightweight package (only dependency is `utils`) and maintained by the R Core Team. My preference is to not try to mimic behavior, but rather rely on the work of the R Core Team to provide these newer functions for older versions that we want to support.

R/utils_memoised.R
dgkf: The rationale for this is that it's only ever used for testing purposes, so the leading `.` is an indication that the parameter isn't intended for most users (in this case, most developers of the package).

With that perspective in mind, do you still think the order and naming should be changed?

R/utils_memoised.R
dgkf: This option isn't intended to be set by users, but is set only for testing purposes and will always be set to the working directory when tests are run. If you can think of a better way to approach this, I'm open to alternatives and agree that an option isn't ideal. 

https://github.com/pharmaR/riskmetric/blob/84d8810f9ad661db39883578b9cec82c1b3ad883/tests/testthat/setup_mock_web_requests.R#L4

R/utils_memoised.R
dgkf: I just prefer to use the r-project url over a url hosted by a private organization. `repos` is an argument, 

https://github.com/pharmaR/riskmetric/blob/84d8810f9ad661db39883578b9cec82c1b3ad883/R/utils_memoised.R#L76

and defaults to `getOption("repos")` - this is the same option that a user would use to redirect to a different package manager repo. `@CRAN@` is a value that I've seen handled specifically in other packages, but ideally this option is almost always set to a url.

R/utils.R
dgkf: Right now it's unused. In the future, if we want to capture output during metric evaluation and print it out to an audit log, we can use this to reformat the output as it would have appeared if the code was run in an interactive session. 
I'll add in an example of what this might look like, but if you want to try it out you can take a look at

```r
# with the ./riskmetric project directory as your working directory

console_output <- riskmetric:::capture_expr_output({
  # output i's, looping through output type
  for (i in 0:10) {
    f <- switch(i %% 3 + 1L, print, message, warning)
    f(i)
  }
})

print(console_output)
# > for (i in 0:10) {
#   f <- switch(i%%3 + 1L, print, message, warning)
#   f(i)
# }
# [1] 0
# 1
# Warning in eval(expr_quote, env): 2
# [1] 3
# 4
# Warning in eval(expr_quote, env): 5
# [1] 6
# 7
# Warning in eval(expr_quote, env): 8
# [1] 9
# 10
```

You can also check out some of the attributes to explore the information we capture during evaluation. 

It also needs to be called `print.<object class>` so that the print function dispatches to the appropriate function for formatting.

R/utils_memoised.R
elong0527: I see. It goods for me. 

In addition, would the argument `e` required ?

https://github.com/pharmaR/riskmetric/blob/127080d9f4f7d45a6d140d4ee0083d3e7a4521be/R/utils_memoised.R#L25

R/utils_memoised.R
elong0527: Thanks for the clarification.   I am clear of the meaning for `.test` now.

R/utils_memoised.R
dgkf: Yes - it at least needs to be a function that accepts one (or more) arguments. The name isn't meaningful, especially when we just suppress it by returning `NULL`. If an issue is caught during evaluation of the expression in `tryCatch`, this function is called before throwing the error. This is how we suppress the output. As a convention, I usually use `m`, `w` and `e` for messages, warnings and errors respectively:

```r
tryCatch(<some expression>
  message = function(m) <handle messages>,
  warning = function(w) <handle warnings>,
  error = function(e) <handle errors>
)
```

elong0527 (approved):

dgkf:
Thanks both @elimillera & @elong0527 - a big milestone and something that I think will help the project to scale as we grow the number of contributors. Thanks for all your work on this one.

PR #169, opened 2020-10-09
Dev/r cmd check

emilliman5:
This replaces half of #137 by creating an `r_cmd_check` pkg_ref_cache slot and accompanying assessment which counts the number of errors warnings and notes that are produced after running `devtools::check` (equivalent to: `R CMD check`). The ref_cache stores the entire check output, which is a list with some pretty printing (from devtools). This currently only works for `pkg_source` with todos left for `pkg_cran_remote` and `pkg_bioc_remote`


R/assess_r_cmd_check.R
```diff
@@ -0,0 +1,25 @@
...
+             class = "pkg_metric_r_cmd_check")
+}
+
+assess_r_cmd_check.pkg_cran_remote <- function(x, ...) {
+  pkg_metric_todo("Assessment of R CMD check on remote pkg refs is not yet implemented but will be in the future")
+}
+
+assess_r_cmd_check.pkg_bioc_remote <- function(x, ...) {
+  pkg_metric_todo("Assessment of R CMD check on remote pkg refs is not yet implemented but will be in the future")
+}
```
dgkf, line 25: In order for `assess_r_cmd_chec(<pkg_ref>)` to work as a stand-alone function, we need to export all of these methods. As it is now, the assessment will only work if it's called through a higher level function like `pkg_assess`.

```suggestion
#' @export
assess_r_cmd_check.default <- function(x, ...) {
  pkg_metric_na(message = "Source code not available to run R CMD check on.")
}

#' @export
assess_r_cmd_check.pkg_source <- function(x, ...) {
  pkg_metric(sapply(x$r_cmd_check[c("notes","errors","warnings")], length), 
             class = "pkg_metric_r_cmd_check")
}

#' @export
assess_r_cmd_check.pkg_cran_remote <- function(x, ...) {
  pkg_metric_todo("Assessment of R CMD check on remote pkg refs is not yet implemented but will be in the future")
}

#' @export
assess_r_cmd_check.pkg_bioc_remote <- function(x, ...) {
  pkg_metric_todo("Assessment of R CMD check on remote pkg refs is not yet implemented but will be in the future")
}
```

R/pkg_ref_cache_r_cmd_check.R
```diff
@@ -0,0 +1,21 @@
...
+}
+
+pkg_ref_cache.r_cmd_check.default <- function (x, ...) {
+  return(NA)
+}
+
+
+#' @importFrom devtools check
+pkg_ref_cache.r_cmd_check.pkg_source <- function(x, ...){
+  check_results <- check(x$path)
```
dgkf, line 18: I know it's not absolutely necessary, but to keep consistency can we call `devtools::check` just to be explicit to new developers.

```suggestion
  check_results <- devtools::check(x$path)
```

dgkf (changes requested):
Looks great.

Just some minor suggestions before we pull it in.

dgkf:
Looks like our CI is failing here because I forgot to mention that we also need to add `devtools` to our `Imports` section if we are including it as a dependency. I think if we get that added in, then our CI will pass and we can get this pulled in.

elong0527:
@dgkf , could you approve this as well? So we can merge it. Thanks!

dgkf (approved):

elong0527:
@emilliman5, seems there are few conflicts. Would you be able to address it before we merge?

Issue #129, opened 2020-06-19
Add labels to score metrics


andyofsmeg:
This would help automatical labelling in the riskassessment app.  The assess* functions have labels but these are not always applicable for scores.

For example, if the last 30 bug reports is scored, a suitable label might be "A percentage of last 30 BugReports whose status is 'closed'"

elong0527:
require follow up discussion between @dgkf  and @marlycormar

dgkf:
### Honing in on the problem

@marlycormar - Before discussing implementation, lets settle on how we want the information made available to the shiny app. 

For example, if we were to provide a tibble of scores, would a label describing the score be sufficient? If so, then I think we could add a description to each scoring function which could be extracted when the scoring function is applied. 

There is one tricky nuance here. That is that assessments are all scored individually, not as a column, and that the class within a column might vary. For example, using the last 30 bugs assesssment, ideally all assessments in the `pkg_assess` tibble output would have the class `c("pkg_metric_last_30_bugs_status" "pkg_metric")`. However, if one package failed, the assessment result will have the class `c("pkg_metric_error", "pkg_metric", ...)`. This is used so that errors can be scored as their own class, but also means that the scoring function that is used will vary across the assessment results.

So I think we need to do a bit of rethinking about this flow to appropriately accommodate this. 

### Possible Implementation

In the situation where an error occurs during an assessment, eg the last 30 bugs, perhaps we can return an error class as a subclass of the expected assessment class. 

```r
x <- pkg_ref("utils")  # utils has no BugReports defined, so will give an error
a <- pkg_assess(x)

class(a$bug_reports)
# [current]> [1] "pkg_metric_error" "pkg_metric" "simpleError" "error" "condition"
# [suggest]> [1] "pkg_metric_last_30_bugs_status" "pkg_metric" "error"
```

This might be preferred because the pkg_metric_* class would be the same regardless of whether an error is produced. Then we could find the appropriate scoring function (and pull a label attributed to it), but will need to push down the error handling into the scoring function instead of handling it at dispatch. Perhaps something like this might work:

```r
metric_score <- function(x, ...) {
  if (inherits(x, "error")) return(<apply error handler>)
  UseMethod(metric_score)
}

metric_score.pkg_metric_last_30_bugs_status <- function(x, ...) {
  mean(x, na.rm = TRUE)
}

attr(metric_score.pkg_metric_last_30_bugs_status, "label") <- 
[...]

PR #173, opened 2020-10-13
updated metric error/na/todo such that we can still propegate scoring labels

dgkf:
- updated the `pkg_metric_error`, `pkg_metric_na` and `pkg_metric_todo` to be subclasses of `pkg_metric_condition` (to easier catch them all as one class for special treatment)
- added util function `firstS3method` (inspired by `getS3method`, but finding returning the first valid method for an object's subclasses)
- added labels for all `metric_score.pkg_metric_*` functions
- added `pkg_metric_eval` as a way of encapsulating evaluation of metrics and allow us to capture errors

Now we should have label attributes for all columns in a scored pkg_ref tibble, and all items in a scored pkg_ref list.


R/assess_downloads.R
```diff
@@ -19,15 +19,16 @@ attr(assess_downloads_1yr, "label") <- "number of downloads in the past year"
...
+    sum(x$downloads$count)
+  })
 }
 
 
 
 #' Defining an Assessment Scoring Function
 #' Score a package for the number of downloads in the past year
-#'
+#'regularized
```
dgkf, line 31: Whoops - looks like I had a stray cursor at some point and typed into this comment. Will revert.

R/pkg_metric.R
```diff
@@ -8,14 +8,61 @@
...
+#'   \code{pkg_metric}
+#' @param env An environment in which \code{expr} is to be evaluated
+#' @inheritParams pkg_metric
+#' 
+#' @return a \code{pkg_metric} object containing the result of \code{expr} 
+#'
+pkg_metric_eval <- function(expr, ..., class = c(), env = parent.frame()) {
+  out <- capture_expr_output(substitute(expr), env = env, quoted = TRUE)
+  out_metric <- as_pkg_metric(out, class = class)
+  if (inherits(out, "error")) out_metric <- as_pkg_metric_error(out_metric)
```
elimillera, line 60: Could we strip out the '.recording' attribute for erroring metrics?
```
> attr(assess_$downloads_1yr, ".recording")
$expr
{
    sum(x$downloads$count)
}

$attributes
NULL

$visible
[1] TRUE

$traceback
NULL

$output
character(0)
```

R/utils.R
```diff
@@ -161,19 +193,30 @@ capture_expr_output <- function(expr, split = FALSE, env = parent.frame(),
...
 
   structure(
-    if (!quoted) expr_quote else expr,
-    value = res$value,
-    visible = res$visible,
-    traceback = cnds_err_traceback,
-    output = outputs[nzchar(outputs)],
-    class = c("expr_output", "expression"))
+    res$value,
+    .recording = list(
```
elimillera, line 197: What does this '.recording' value do?

elimillera (approved):
Looks great. I was thinking of a different way to do this but this works a lot better now that I've had a look at it.

R/utils.R
dgkf: .recording is a list of console output during the evaluation of the expression. Each line of stdout (from `cat`, `print`) is its own entry in the list, and each `message`, `warning`, or `error` gets its own entry. The print method prints these out as though we're playing back a recording of the execution in an interactive session. 

For now this is just used to capture errors, but we could use something like this to produce a trace of the steps that are executed to collect all the pkg_metric info if we want to generate a report.

I'm not sold on the complexity here, so if it ends up being problematic we can simplify it.

marlycormar:
I second @elimillera's comment =) 

This PR works towards issue #129.

PR #185, opened 2021-02-03
Update pkg_ref_cache_function_coverage.R

elong0527:


dgkf (changes requested):

elong0527:
@dgkf, After this change, the code is exactly the same as in `https://github.com/pharmaR/riskmetric/blob/master/R/pkg_ref_cache_expr_coverage.R` 

Could we just remove `pkg_ref_cache_function_coverage.R` ?

dgkf:
Ah, I didn't realize we had both. That sounds good, let's get rid of this function then.

elong0527:
opened a new pull request #188 to address conflicts

Issue #182, opened 2021-01-20
riskmetric v0.1.0 CRAN release check list 


elong0527:
- [x] double check package dependency
- [x] Polish NEWS
- [x] Polish pkgdown reference index
- [x] confirm package author list 
- [x] confirm package version 
- [x] devtools::check()
- [x] devtools::check_win_devel()
- [x] rhub::check_for_cran()
- [x] add cran-comments.md file.

dgkf:
Adding feedback from CRAN submission automated checks:

```
Dear maintainer,

package riskmetric_0.1.0.tar.gz does not pass the incoming checks automatically, please see the following pre-tests:
Windows: <https://win-builder.r-project.org/incoming_pretest/riskmetric_0.1.0_20210405_211039/Windows/00check.log>
Status: 2 NOTEs
Debian: <https://win-builder.r-project.org/incoming_pretest/riskmetric_0.1.0_20210405_211039/Debian/00check.log>
Status: 2 NOTEs



Please fix all problems and resubmit a fixed version via the webform.
If you are not sure how to fix the problems shown, please ask for help on the R-package-devel mailing list:
<https://stat.ethz.ch/mailman/listinfo/r-package-devel>
If you are fairly certain the rejection is a false positive, please reply-all to this message and explain.

More details are given in the directory:
<https://win-builder.r-project.org/incoming_pretest/riskmetric_0.1.0_20210405_211039/>
The files will be removed after roughly 7 days.

No strong reverse dependencies to be checked.

Best regards,
CRAN teams' auto-check service
Flavor: r-devel-windows-ix86+x86_64
Check: CRAN incoming feasibility, Result: NOTE
  Maintainer: 'Doug Kelkhoff <doug.kelkhoff@gmail.com>'

  New submission

  Found the following \keyword or \concept entries with Rd markup:
    File 'assess_covr_coverage.Rd':
      \concept{\code{assess_*} functions}
    File 'assess_downloads_1yr.Rd':
      \concept{\code{assess_*} functions}
    File 'assess_export_help.Rd':
      \concept{\code{assess_*} functions}
    File 'assess_has_bug_reports_url.Rd':
      \concept{\code{assess_*} functions}
    File 'assess_has_maintainer.Rd':
      \concept{\code{assess_*} functions}
    File 'assess_has_news.Rd':
      \concept{\code{assess_*} functions}
    File 'assess_has_source_control.Rd':
      \concept{\code{assess_*} functions}
[...]

PR #201, opened 2021-04-18
address CRAN submission comments

elong0527:
Addressed CRAN comments discussed in #182


elong0527:
Every Rd file now contain a `\value` field after checking using the code below.

```
devtools::document()

path <- "man"
file <- sort(list.files(path, full.names = TRUE))

res <- lapply(file, function(x){
  value = grep("\\value", readLines(x), fixed = TRUE, value = TRUE)
  data.frame(file = x, value = ifelse(is.null(value), NA, value))
})

res <- do.call(rbind, res)

View(res)
```

cran-comments.md
```diff
@@ -0,0 +1,14 @@
...
+
+## R CMD check results
+
+0 errors | 0 warnings | 0 note
+
+## Response from last CRAN submission comments
+
+- There is no existing references describing the methods in this package. 
+- Added return \value item to Rd
+- Removed examples of unexported functions.
```
dgkf, line 14: Copy and pasting some content from the reviewer in case the reviewer for 
next submission cycle doesn't have this context.

```suggestion
> If there are references describing the methods in your package, please
> add these in the description field of your DESCRIPTION file [...]

- There is no existing references describing the methods in this package. 

> Please add \value to .Rd files regarding exported methods and explain
> the functions results in the documentation. Please write about the
> structure of the output (class) and also what the output means. [...]

- Added return \value item to Rd

> You have examples for unexported functions.
> Please either omit these examples or export these functions. [...]

- Removed examples of unexported functions. 
```

dgkf (approved):
Thanks for going through and adding all of these, @elong0527. I just added in some small quotes from the reviewer email for cran-comments context.

PR #205, opened 2021-05-21
Update workflow image path and location

### LINKED ISSUE #204: missing image from packagedown site
image

elimillera:
Thanks to @bms63 for finding this!

I copied the workflow image to man/figures which I believe is the standard place to put images in vignettes. Updated the fig.path in the vignette and linked to that instead.

Closes #204


dgkf (approved):
Thanks for getting this added back in @elimillera! Tested with a fresh build and the image gets loaded into the knit'ed html file.

elong0527 (approved):

Issue #207, opened 2021-07-06
Class hierarchy of pkg_remote isn't maintained


dgkf:
Thanks to a bug report, #206 - it looks like our recent overhaul of the class constructors (#203) dropped some of the intermediate classes needed for metric inheritance.

Using the example in that bug report,

```r
pkg_visR <- pkg_ref("visR")  # when not installed
class(pkg_visR)
#> # produces:
#> c("pkg_cran_remote", "pkg_ref", "environment") 
#> # expect:
#> c("pkg_cran_remote", "pkg_remote", "pkg_ref", "environment") 
```

As described in `?pkg_ref`, we expect that `pkg_remote` is a superclass of `pkg_cran_remote`, `pkg_bioc_remote` and `pkg_git_remote`. 

Because of this, many metrics for cran and biocondcutor remotes fail to inherit generic metrics implemented for `pkg_remote`s.

PR #208, opened 2021-07-06
Fix an oversight in classes applied to pkg_ref when source is explicitly provided

### LINKED ISSUE #206: Data the score is calculated on differs based on whether the package is installed locally or not
I tried to automatically scrape all of CRAN for riskmetric scores and quickly noticed that the results massively favoured locally installed packages. After some more digging, it seems that, when pulling from CRAN instead of local, many of the metrics fail.  

Is this behaviour local to my machine?

image
image

dgkf:
As reported in #206 & isolated in #207, a user discovered an issue where `pkg_remote` methods weren't being dispatched upon because this superclass was omitted from `pkg_*_remote` subclasses. 

I added in a programmatically accessible class hierarchy (nested list) which is traversed in search of valid `source` classes, and applies the entire class hierarchy to the resulting object. eg

```r
class(pkg_ref("mypkg", source = "pkg_cran_remote"))
#> c("pkg_cran_remote", "pkg_remote", "pkg_ref", "environment")

class(pkg_ref("mypkg", source = "pkg_install"))
#> c("pkg_install", "pkg_ref", "environment")
```

Only concern I have with this approach is allowing users to extend with their own subclasses, but we can cross that bridge when we get there. I don't see any demand for that functionality quite yet. 

@elimillera - would you mind taking a look since most of the changes affect code that was introduced in #203?

Closes #206, #207

---

Since this is actually a fairly crippling bug for remote package refs, I think this might warrant a resubmission to CRAN for the patch release.


elimillera:
Thanks @dgkf , I thought I was being smart by simplifying the classes of the pkg_ref objects. This looks good to me

elimillera (approved):

PR #221, opened 2021-09-01
Dev/cohort

emilliman5:
This the start of a cohort-level analysis of risk. This is a very rough draft, but I am hoping it will seed a number of discussions on how we implement a cohort class, how we dispatch methods to it, what a cohort metric is and is not.

To do: 
1. More robust cohort_ref constructors
  1. should be able to handle different types of libraries (installed, remote, etc)
  2. should make assessment of the cohort against the library easy to implement (right now it is a bit convoluted and ugly)
  3. Maybe add a meta/aggregate ref slot for the library
2. Better defintions of what goes into the `ref` object vs `assessment` object
3. A more formal cohort assessment class structure; currently just using a list
4. methods to handle the case where library is `NULL` to do a cohort comparison against itself.


R/assess_reverse_dependencies.R
```diff
@@ -0,0 +1,48 @@
...
+  pkg_metric_eval(class = "pkg_metric_reverse_dependencies",
+                  devtools::revdep(x$name, bioconductor = TRUE)
+  )
+}
+
+#' @importFrom devtools revdep
+#' @export
+assess_reverse_dependencies.cohort_ref <- function(x, ...){
+  #cohort_rev_deps <- devtools::revdep(x$cohort$name, bioconductor = TRUE)
+  cohort_rev_deps <- lapply(x$cohort, assess_reverse_dependencies)
```
dgkf, line 30: Super weird S3 dispatch within an apply - try this:

```suggestion
  cohort_rev_deps <- lapply(x$cohort, function(i) assess_reverse_dependencies(i))
```

emilliman5:
```
cohort_rev_deps <-x$cohort %>% assess_reverse_dependencies()

```
or

```
cohort_rev_deps <-x$cohort %>% pkg_assess( assessments = assess_reverse_dependencies())
```

R/assess_dependencies.R
```diff
@@ -0,0 +1,107 @@
...
+
+#' Gets available packages from necessary repository and filters for
+#' package of interest
+#'
+#' @param name package name
+#' @param repo package repository (e.g. CRAN or Bioconductor)
+#'
+#' @return Returns a data frame with two columns 1) Package names, 2) type of dependency (LinkingTo, Imports, Depends)
+#'
+get_package_dependencies <- function(name, repo){
```
daroczig, line 80: does this need to be recursive to get the actual footprint? like `miniCRAN::pkgDep` does:

> Performs recursive retrieve for ‘Depends’, ‘Imports’ and ‘LinkLibrary’. Performs non-recursive retrieve for ‘Suggests’.

R/cohort_ref_class.R
```diff
@@ -0,0 +1,42 @@
...
+  if (missing(x)) return(structure(list(pkg_ref_list = list(logical(0L)),
+                              library = logical(0L)), class = "cohort_ref"))
+  new_cohort_ref(x, library, ...)
+}
+
+new_cohort_ref <- function(x, library, ...){
+  dots <- list(...)
+  if (length(dots) && is.null(names(dots)) || any(names(dots) == ""))
+    stop("cohort_ref ellipses arguments must be named")
+  if(is.character(x) | all(sapply(x, class) != "pkg_ref")){
```
daroczig, line 19: Sorry, this is really offtopic, but seeing two different code styles for the `if` conditions (e.g. space VS no space after `if`, and requiring explicit curly braces even for atomic expressions VS skipping the optional curly brace in that case), I wonder if there are any plans to introduce `lintr` in the PR checks and enforce a common style?

dgkf: It wasn't on our roadmap, but I'm not opposed to adding it in. I generally find `lintr` to be a bit too opinionated and lacks a lot of the core functionality that makes other language linters/formatters useful (indentation, function header/call arg formatting, line-width aware linting).

Some of the lintrs that end up being particularly problematic are function name length (since it doesn't account for S3 class names used in dispatch) and camelCase/snake_case linters that throw lints for a lot of imported base functions.

I'd be fine adding in a selective set of linters for less-opinionated things like whitespace and line length.

R/assess_dependencies.R
```diff
@@ -0,0 +1,107 @@
...
+}
+
+#' Parse DCF of description file
+#'
+#' @param path pkg_ref path
+#'
+parse_dcf_dependencies <- function(path){
+  dcf <- read.dcf(file.path(path, "DESCRIPTION"), all=TRUE)
+  dcf <- dcf[colnames(dcf) %in% c("LinkingTo","Imports", "Depends")]
+  dcf <- sapply(dcf, strsplit, strsplit, split=",")
```
dgkf, line 99: Be careful with dependencies that include a version constraint (like `dplyr (>= 1.0.0)`)

R/cohort_ref_class.R
daroczig: Yeah, I agree that `lintr` is far from perfect, and sometimes really annoying -- but I've found it very useful in collaborative projects to keep the code style/whitespace etc consistent and minimize the related (sometimes very opinionated) discussions once it's enforced.

E.g. setting custom line and object length, disabling some others (e.g. I'm a single-quote person) etc.

And still heavy use of `# nolint` for false positives, but at least that's explicit.

R/assess_reverse_dependencies.R
dgkf: I don't think we ever want to use `pkg_assess` on a cohort directly. Instead, maybe we need a `cohort_assess` helper that applies all available package assessments (likely using `pkg_assess` under the hood), but also including context and cohort metrics. 

Let's take two examples, a generic "context" assessment `assess_context_metric` and a generic "cohort" assessment `assess_cohort_metric`. Just to recap (and make sure I'm remembering correctly), a context metric is a per-package metric that is contextualized by a cohort, and a cohort metric is a per-cohort metric (producing a single score for an entire cohort). 

Then we'd have these scenarios

### Given a cohort, calculate a specific _context_ metric for _a single_ package

```
# just mocking up some example params, feel free to suggest alternatives
assess_reverse_dependencies(<pkg>, context = <cohort>, ...) -> <cohort_of_pkg_metric (len 1)>
```

### Given a cohort, calculate a specific _context_ metric for _all_ cohort packages

```
assess_reverse_dependencies(<cohort>, ...) -> <cohort_of_pkg_metric (len n)>
```

### Given a cohort, calculate a single _cohort_ metric

```
assess_cohort_metric(<cohort>, ...) -> <cohort_metric>
```

### Given a cohort, calculate all _context_ & _cohort_ metrics for _all_ cohort packages

Analogous to how `pkg_assess` is a convenience function for calculating all available assessments for a `list_of_pkg_ref` objects, we need a way to run all the available cohort-style metrics on a cohort. The return type should be similar to `pkg_assess` (one metric per assessment per package), but with additional cohort metrics (one metric per cohort metric)

```
cohort_assess(<cohort>, ...)
```

emilliman5:
closing for now. Will create new PR's more modular fashion. First, cohort_ref constructor, then a few assessments, then some metrics, then a score

Issue #218, opened 2021-08-24
riskmetric is archived on CRAN


llrs:
I wanted to look at the package but I found it is no longer on [CRAN](https://cran.r-project.org/package=riskmetric):

> Package ‘riskmetric’ was removed from the CRAN repository.
> 
> Formerly available versions can be obtained from the [archive](https://cran.r-project.org/src/contrib/Archive/riskmetric).
> 
> Archived on 2021-08-17 as repeated check problems were not corrected in time.
> 
> A summary of the most recent check results can be obtained from the [check results archive](https://cran-archive.r-project.org/web/checks/2021/2021-08-17_check_results_riskmetric.html).
> 
> Please use the canonical form [https://CRAN.R-project.org/package=riskmetric](https://CRAN.R-project.org/package=riskmetric) to link to this page.
> 

The only check I found failing is on r-devel-fedora-gcc (copying it here as I don't know how long these checks archives stay):

<details>
<summary>Check Details</summary>

> Version: 0.1.1
> Check: tests
> Result: ERROR
>      Running ‘testthat.R’ [150s/168s]
>     Running the tests in ‘tests/testthat.R’ failed.
>     Complete output:
>      > library(testthat)
>      > library(riskmetric)
>      >
>      > options(repos = "fake-cran.fake-r-project.org")
>      >
>      > test_check("riskmetric")
>      ══ Skipped tests ═══════════════════════════════════════════════════════════════
>      • snapshots appear to be failing due to random order of pkg_ref fields (3)
>     
>      ══ Failed tests ════════════════════════════════════════════════════════════════
>      ── Failure (test_pkg_ref.R:31:3): pkg_ref infers source when not explicitly provided ──
>      `ref3` has class 'pkg_install'/'pkg_ref'/'environment', not 'pkg_bioc_remote'/'pkg_remote'/'pkg_ref'/'environment'.
>      ── Failure (test_pkg_ref.R:32:3): pkg_ref infers source when not explicitly provided ──
>      ref3$source (`actual`) not equal to "pkg_bioc_remote" (`expected`).
>     
>      `actual`: "pkg_install"
>      `expected`: "pkg_bioc_remote"
>     
>      [ FAIL 2 | WARN 0 | SKIP 3 | PASS 45 ]
>      Error: Test failures
>      Execution halted
> Flavor: r-devel-linux-x86_64-fedora-gcc

</details>

Not sure if you were aware, but is there a plan to submit the package to CRAN again?
[...]

PR #223, opened 2021-09-15
Switch tests to use source functions

elimillera:
Per the CRAN archival #218 , its unclear what bioc packages are on the VMs CRAN is testing on. This makes the calls explict.

WIP. Looking to add in @dgkf suggestions.


elimillera:
@dgkf So the tests are passing but its still expecting the ubuntu 16 tests to be passing for some reason

dgkf (approved):
lgtm, and passing on 18.04 even if the 16.04 runners are inaccessible

PR #226, opened 2021-10-01
Updating GitHub actions, using more recent r-lib actions

### LINKED ISSUE #225: Update GHA
Realized while debugging our CI scripts that our builds are a bit outdated (still using Ubuntu 16.04LTS). I'm sure other components are outdated as well. 

Let's update to a more recent version of the actions

dgkf:
- Clear out existing actions
- Added new (equivalent, but updated) actions using `usethis::use_github_action*` helpers

Closes #225


dgkf:
@elong0527 - tagging you for review since I think it was you who added the first round of github actions. Looks like updating the ubuntu images fixes our failing runs.

PR #234, opened 2021-11-15
Dev/dependencies metric

emilliman5:
New metric on a package's dependency foot print. Also a helper function to pass specific assessments to `pkg_assess` which aids the inspection of how an assessment is returned by `pkg_assess`


R/assess_dependencies.R
```diff
@@ -0,0 +1,121 @@
...
+assess_dependencies.pkg_install <- function(x, ...){
+  pkg_metric_eval(class = "pkg_metric_dependencies", {
+    parse_dcf_dependencies(x$path)
+  })
+}
+
+#' @export
+assess_dependencies.pkg_cran_remote <- function(x, ...){
+  pkg_metric_eval(class = "pkg_metric_dependencies", {
+      get_package_dependencies(x$name, repo = getOption("repos")[["CRAN"]])
```
dgkf, line 43: I don't think we necessarily need to change this before pulling in changes, but we may also want to consider unnamed repos that match a repo listed in `utils::getCRANmirrors()` (we already cache this with `memoise_cran_mirrors()`) so that we don't require that `getOption("repos")` always have names.

R/assess_dependencies.R
```diff
@@ -0,0 +1,121 @@
...
+#' Helper function to remove base and recommended packages
+#'
+#' @param df Data frame of dependencies of a package.
+#'
+remove_base_packages <- function(df){
+  deps <- df[!grepl("^R\\s\\(.+\\)", df$package) | df$package %in% c('base','compiler','datasets',
+                                                                           'graphics','grDevices','grid',
+                                                                           'methods','parallel','splines',
+                                                                           'stats','stats4','tcltk','tools',
+                                                                           'utils'), ] ##Remove base R dependencies and in the future recommended pacakges
```
dgkf, line 119: Instead of hardcoding, we can use `installed.packages()[,"Priority"]` to get a list of base/recommended packages:

```r
inst <- memoise_installed_packages()
inst_prio <- inst[,"Priority"]
inst_is_base_rec <- !is.na(inst_prio) & inst_prio %in% c("base", "recommended")
base_rec_pkgs <- inst[inst_is_base_rec, "Package"]
```

dgkf (approved):
Looks good - just a few suggestions, but nothing that would drastically change behavior, just maintainability. I'll leave it up to you whether you want to incorporate the feedback now or would prefer that it is slated as a separate body of work.

R/assess_dependencies.R
emilliman5: incorporated this suggestion, ended up taking it one step further by checking against memoise_cran_mirrors, if that returns empty then checking all urls in getOptions("repos") for a "cran" subdomain and finally for a named "CRAN" repo

PR #243, opened 2022-02-18
Issue 241

### LINKED ISSUE #241: pkg_score Can it be setup to ignore cran checks?
I attempted to run the package functions to get the score of RemixAutoML. Unfortunately, it runs a cran check and mine will fail every time, by design. 

I setup the package to only include the necessary packages, needed to install and use, in the DESCRIPTION file and all others are left as optional to the user, depending on which functions they'd like to use. For example, I have some functions that utilize the package 'h2o' but if a user doesn't intend to use any of them, then they won't have to install h2o or the java sdk. I believe that is a functionality that is desirable, but cran does not agree. Nonetheless, I'd love to test out the functions here but I won't be able to get a score because I'll will get an error on the cran checks.

Thanks for reading this far if you did :)

emilliman5:
added some functions and tests for more convenient setting of metric weights and selecting assessments to pass to `pkg_assess`. These were sparked by #242


R/pkg_assess.R
```diff
@@ -92,19 +92,39 @@ all_assessments <- function() {
...
-get_assessments <- function(fxn_string=""){
+get_assessments <- function(fxn_string="", invert=FALSE){
+  fs <- grep("^assess_[^.]*$",
+             getNamespaceExports(utils::packageName()),
+             value = TRUE)
+  fxn_string <- lapply(fxn_string, grep, x=fs, value=T)
+
+  if(length(fxn_string)==0 | all(sapply(fxn_string, length)==0)){
+    warning("Did not find any assessments that matched your query. Returning all assessments")
+    return(all_assessments())
```
elimillera, line 109: If I'm reading this right it will return all assessments if you pass a string that doesn't grep for any of the assessments. I'm thinking it might be better to just return nothing if fxn_string doesn't find anything. I'm thinking it might be unexpected behavior if we return something that wasn't actually found by group

R/pkg_assess.R
```diff
@@ -92,19 +92,39 @@ all_assessments <- function() {
...
+  if(invert){
+    fxn_string <- fs[!fs %in% fxn_string]
+  }
+
+  Map(getExportedValue,
+      fxn_string,
+      ns = list("riskmetric"))
   Map(getExportedValue,
       fxn_string,
       ns = list(utils::packageName()))
```
elimillera, line 125: These look like they're duplicated.

PR #249, opened 2022-06-07
Adding the function of assessing number of lines of code base

shengwei66:
I have created a draft function, assess_size_codebase, for the pkg_install class.


R/assess_size_codebase.R
```diff
@@ -0,0 +1,27 @@
...
+}
+
+attributes(assess_size_codebase)$column_name <- "size_codebase"
+attributes(assess_size_codebase)$label <- "number of lines of code base"
+
+#' @export
+assess_size_codebase.pkg_install <- function(x, ...) {
+  pkg_metric_eval(class = "pkg_metric_size_codebase", {
+    # create character vector of exported function
+    exports <- paste0(x$name,"::", getNamespaceExports(x$name))
```
emilliman5, line 19: What about non-exported functions of a package, i.e. ":::".

emilliman5 (changes requested):
This is a great start

emilliman5:
There is a problem when a package exports a function with special characters (e.g. %>%, %+%, etc). Can you investigate ways to parse-eval-captures strings with special characters? At the very least the assessment will have to error/warn gracefully if there is no solution to this, (i.e. tryCatch()).

Here is a reproducible example, assuming ggplot2 is an installed package:

```
pr <- pkg_ref("ggplot2")
pa <- pr %>% pkg_assess()
pa$size_codebase
#<simpleError in str2lang(x): <text>:1:10: unexpected SPECIAL
#1: ggplot2::%+replace%
#             ^>
```

mpadge:
My proverbial 2-cents-worth: There are a lot of R packages that don't really have code in the R directory at all. A great example comes from one of the most talented coders i've ever worked with, who only writes a couple of wrapper lines in R at the end of a heap of coding in various other directories. This is what comes from using the approach suggested here to analyse one of his packages:
``` r
path <- file.path (tempdir (), "humanleague")
gert::git_clone ("https://github.com/virgesmith/humanleague", path = path)
devtools::load_all (path, export_all = FALSE)
#> ℹ Loading humanleague

exports <- paste0("humanleague::", getNamespaceExports("humanleague"))
nloc <- sapply(exports, function(x){length(capture.output(eval(str2lang(x))))})
sum(nloc) - 2*length(nloc)
#> [1] 16
```
And that will suggest the package has virtually no code at all. It is in fact enormous, as revealed by this:
``` r
pkgstats::loc_stats (path)
#> # A tibble: 5 × 12
#> # Groups:   language, dir [5]
#>   language dir   nfiles nlines ncode  ndoc nempty nspaces nchars nexpr ntabs
#>   <chr>    <chr>  <int>  <int> <int> <int>  <int>   <int>  <int> <dbl> <int>
#> 1 C        src        1     34    28     0      6     153   1077   1       0
#> 2 C++      src       19   3376  2063   756    557    8708  77340   1      15
#> 3 Python   tests      1    335   240    30     65    1082   8617   2       0
#> 4 R        R          2    202    25   164     13     999   7185   1       0
#> 5 R        tests      2    680   457    93    130    1525  16972   1.5     0
#> # … with 1 more variable: indentation <int>
```

<sup>Created on 2022-06-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

This gives the correct value of 25 lines-of-code in the R directory, yet there are nearly one hundred times that again in the `src` directory. (Plus as an additional note, `eval` is potentially dangerous here, because you don't know whether all objects will indeed be `eval`-uable. It'd be much safer to drop the `"<package>::"` prefix, and just use `get(str2lang(x), envir=as.environment("package:<package>"))`. Also far more efficient.)

---

NOTE: I do no mean this to be any kind of advertising for using the `pkgstats` package, and have no expectation at all that it actually be used here. Nevertheless, if you do want to consider it, it is now on CRAN, so easier to use as a straight dependency here. That said, cross-platform installation remains non-trivial, so ... happy to discuss if you do decide to try it. It'll definitely give you the easiest one-line drop-in function for doing what you want here.

shengwei66:
Thanks for @mpadge comments. I have revised the assess_size_codebase function by using the mget function instead of eval. @emilliman5 after implementing the mget function, I am able to run through the ggplot2 example which has the special character function.

``` r
pr <- pkg_ref("ggplot2")
pa <- pr %>% pkg_assess()
pa$size_codebase
#> [1] 6153
```

<sup>Created on 2022-07-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

elimillera (commented):
I think the setup for pkg_install is really good. @shengwei66 would it be possible to write one for pkg_source as well? I think that would be straight-forward. And I think we could add a suggests for pkgstats for pulling the data for pkg_remote.

shengwei66:
@elimillera and @emilliman5, can you please provide me an example package that will have the class of pkg_source? So I can use the example to develop the function of calculating the size of codebase. Thanks.

emilliman5 (approved):

shengwei66:
@emilliman5 and @elimillera, I have added the function for the pkg_source class. Can you please review and let me know if any comments? Thanks.

PR #251, opened 2022-06-23
Adds install.packages() instruction to readme

eliocamp:
The package is now on CRAN, so the readme was outdated.


elimillera (approved):
Looks good. Thanks for the update!

Issue #210, opened 2021-07-13
Integrations with `oysteR`


dgkf:
Thanks @csgillespie for reaching out at useR! - This looks like a really cool package to integrate with. 

[`oysteR`](https://github.com/sonatype-nexus-community/oysteR/) looks like an awesome resource for identifying security vulnerabilities in packages. From a brief readthrough, it looks like it compares a package against the [Sonatype OSS Index](https://ossindex.sonatype.org/), identifying known vulnerabilities.

csgillespie:
After our slack chat, I realised that the current functions in {oysteR} rely on the pkg being installed, which is a little silly. So that needs to be fixed first (quite easy).

dgkf:
Not a problem - some of our metrics also hinge on having the package installed (`R CMD check`, `coverage`), but each of those manage installation themselves. It would definitely streamline things if it's not a prerequisite.

Issue #209, opened 2021-07-13
How to integrate metrics from other packages?


dgkf:
As the audience of `riskmetric` grows, we're seeing new packages that offer really compelling metrics, but might introduce new dependencies, drastically increase runtime or only apply in niche applications. One of our design goals has always been extensibility, yet we don't have any concrete examples of how a package might extend riskmetric. 

I've added the [Package Integration](https://github.com/pharmaR/riskmetric/labels/Package%20Integration) label to track our backlog specific to packages which may want to register metrics with `riskmetric`. 

### Option 1: All metrics created in `riskmetric`

Perhaps the most straightforward solution is just to implement metrics within `riskmetric`. However, this would bloat our dependency footprint and runtimes meaning that we may want to build more interfaces to restricting which sets of metrics get run. 

### Option 2: All metrics created in `riskmetric`, but only run if `Suggests` dependency is installed

As a way of mitigating this management, perhaps we could only turn on these behaviors when those dependencies are available. However, this feels like an enormous amount of dependency management, which really should be handled by the package installation instead.

### Option 3: Offshoot packages (eg, `riskmetric.<use case>`)

Alternatively, we could develop more focused packages - perhaps by theme such as `riskmetric.security`, or by package that is integrated such as `riskmetric.srr`or `riskmetric.autotest`. This would let us fragment metrics among a few different extension packages so that specific metric classes or packages handle all the bookkeeping for their own dependencies.

### Option 4: Encourage packages to refer to `riskmetric` via `Enhances`

I think the devil might be in the details with this one, but we could aim to make it as easy as possible for other packages to register metrics that enhance `riskmetric` - possibly without even taking on a direct dependency since all of our extensions operate through S3 dispatch. I think this could be the most comfortable way to encourage extensions and we could rely on a canonical way of tracking enhancement via the DESCRIPTION `Enhances` field. I think a proof-of-concept would be needed to make sure this is feasible before encouraging this path forward, though. 

### Option ... ? 

Open to suggestions.

## Action Items

Whatever path we take, I think we could use a new vignette showing how to integrate new metrics that bring niche or intensive metrics to `riskmetric`.

PR #272, opened 2023-02-03
Add security assessment via {oysteR}

borgmaan:
This is a draft of what the integration with `oysteR` could look like (#210). I am looking for feedback on some design decisions for following the `Suggests` integration approach for metrics from other packages (Option 2 in #209). Hopefully this draft PR is a good place to raise these questions and gather feedback. Please advise if another forum is preferred. 

**********

### 1) Handling calls to `assess_*` functions when package from `Suggests` is not present
I added [some logic](https://github.com/pharmaR/riskmetric/blob/be632831f5a41ed48c1e72741c529a165b2e04a5/R/assess_security.R#L11-L32) to the generic function to detect whether the `Suggests` dependency is available and prompt the user to install the dependency if it is not. The package is installed and the method dispatch continues as expected if the user chooses to install. An error is raised if the installation is not performed. 

This seemed like reasonable behavior to me but I welcome feedback on what the end-user experience should look like here. Also, I noticed custom exceptions/conditions being used throughout the package. Is this a situation where we would want to raise something other than the default from `stop`?

### 2) Hooks to `all_assessments`-based workflows
Another area to address when extending the package in this manner is the behavior of the `all_assessments` function. This currently works by `grep`-ing the packages exported functions to find those starting with `assess_*`. 

I have [modified this mechanism](https://github.com/pharmaR/riskmetric/compare/master..dev/security#diff-e2b027c2af6c76ef45f38cce79a1549ea69636e7983ebd6239b6e914f0e88803) to allow `assess_*` functions with dependencies present in `Suggests` to be excluded from `all_assessments()` based on the [presence of an attribute](https://github.com/pharmaR/riskmetric/blob/be632831f5a41ed48c1e72741c529a165b2e04a5/R/assess_security.R#L39). The full list of `assess_*` objects can be returned by passing `include_suggests = TRUE`).

This seemed in line with the desire to not impose additional dependencies on the end-user but I think there may be many ways to accomplish this (e.g. different naming conventions, etc.). I am not sure what is best.

What are others thoughts on what the behavior should look like here?

### 3) `pkg_ref_cache.security.default` Implementation 
This is less integration focused and more just a question on implementing this metric. The [current implementation](https://github.com/pharmaR/riskmetric/blob/1b1eebece25dbf2c1be55237369315620371c452/R/pkg_ref_cache_security.R#L16-L37) does the following:

1. Populates the cache of dependencies by calling `assess_dependencies()`.
2. Build a new list of `pkg_refs` including the original package and its dependencies.
3. Uses this list to get vectors of package names and versions to pass to `oysteR::audit`.

This seemed like a good way to get the package dependencies and versions that were relevant to the "assessment context". It also made it easy to cover a number of `pkg_ref` sources (source, cran, bioconductor) in a single `default` method. 

Admittedly, I am still building my mental model of all the package internals. Is this seem like a good way to accomplish this? Anything to look out for?

*****

There is still work to be done to clean this up and build out supporting tests but was hoping to get some feedback at this early stage. @elimillera @emilliman5 looking forward to your feedback on the above points. And are there others to include in the design discussion?


borgmaan:
- how to make prompts for package install work smoothly when you are passing multiple `pkg_ref` objects? only prompt once? or store choice?
- update [this](https://github.com/pharmaR/riskmetric/blob/1b1eebece25dbf2c1be55237369315620371c452/R/pkg_ref_cache_security.R#L16-L37) to just be specific to the single package under assessment
    - can re-use this implementation in the future to assess security for a cohort?
- is there a way to have `all_assessments()` return `Suggests` assessments if the package is present and not if it is not installed?

R/assess_security.R
```diff
@@ -0,0 +1,62 @@
...
+#' @param x a \code{pkg_ref} package reference object
+#' @param ... additional arguments passed on to S3 methods, rarely used
+#' @return a \code{pkg_metric} containing Assess for any known security vulnerabilities in the OSS Index via oysteR
+#' @seealso \code{\link{metric_score.pkg_metric_security}}
+#'
+#' @importFrom utils install.packages menu
+#' @export
+assess_security <- function(x, ...) {
+  # TODO: discuss preferred approach for handling packages within Suggests
+  if (!requireNamespace("oysteR", quietly = TRUE)) {
```
parmsam-pfizer, line 12: Does the version of oysteR they have installed matter?

borgmaan: > Does the version of oysteR they have installed matter?

Good catch. Addressed in 1a7e7da

> Also, more of a general question is if we need to create some tests for each new metric.

A good discussion to have generally. I did plan on adding them for this metric once the `Suggests` implementation details were sorted out.

borgmaan:
Updates based on feedback from 2/8 sprint meeting:

> * How to make prompts for package install work smoothly when you are passing multiple `pkg_ref` objects? Only prompt once? Or store choice?

Handled via a package-level option that will suppress further attempts to install the package (see [07478e7](https://github.com/pharmaR/riskmetric/pull/272/commits/07478e7c7325b93dfb21761f5c4dfab8492fcedb)).


> * Update [this](https://github.com/pharmaR/riskmetric/blob/1b1eebece25dbf2c1be55237369315620371c452/R/pkg_ref_cache_security.R#L16-L37) to just be specific to the single package under assessment

Only the `pkg_ref` package is now assessed and dependencies logic has been removed (see [c81c54a](https://github.com/pharmaR/riskmetric/pull/272/commits/c81c54a6d0c9a0e644373a3d41e9ef07617e2303))

> * Is there a way to have `all_assessments()` return `Suggests` assessments if the package is present and not if it is not installed?

Security metric is now included in `all_assessments` by default if `oysteR` is installed but excluded otherwise (see [0eeafc4](https://github.com/pharmaR/riskmetric/pull/272/commits/0eeafc4780f730c5920fba77c164942c0c9dbcdc))

csgillespie:
If you need anything from {oysteR}, please let me know.

emilliman5:
Testing strategy:

1) Tests to ensure assessments that operate in a plugin manner behave consistently ("API" side tests)
2) Tests of the actual assessments and metrics can be written in a traditional manner. We will added necessary packages to the test environments
3) At least 1 environment should not have plugin packages to test usability in absence of plugin packages.

parmsam-pfizer:
Ran into this error after installing the branch.
``` r
library(riskmetric)

dplyr_pref <- pkg_ref("dplyr")

dplyr_pref$security
#> Error in `dplyr::bind_rows()`:
#> ! Can't combine `version` <character> and `version` <package_version>.
```

<sup>Created on 2023-03-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

borgmaan:
@emilliman5 (and any others) -- this should be ready to review now. I believe I have checked off everything we discussed to add as functionality and have covered the desired testing scenarios. 

Note, the `oldrel-4` CI job is failing as `oysteR` has dependencies that rely on R > 4.0.0 ([details](https://github.com/pharmaR/riskmetric/actions/runs/5269310696/jobs/9527263879#step:5:156)). All other `R CMD check` runs are passing (as is code coverage). 

Also, as a reminder, there is a wiki page that describes the implementation strategy that may be helpful when reviewing: 

https://github.com/pharmaR/riskmetric/wiki/Adding-Metrics-from-External-Packages

R/pkg_ref_cache_security.R
emilliman5, line 1: I think for the ref_cache we should capture the oysteR output, if ony the tibble to start, but the entire message would be best (especially the overview). 

> oysteR::audit("dplyr", version = "1.0.8", type = "cran")
> ℹ Using cached results for 1 package
> 
> ── Vulnerability overview ──
> 
> ℹ 1 package was scanned
> ℹ 1 package was found in the Sonatype database
> ℹ 0 packages had known vulnerabilities
> ℹ A total of 0 known vulnerabilities were identified
> ℹ See https://github.com/sonatype-nexus-community/oysteR/ for details.
> \# A tibble: 1 × 8
>    package version type  oss_package      description reference vulnerabilities no_of_vulnerabi…
>    <chr>   <chr>   <chr> <chr>            <chr>       <chr>     <list>                     <int>
> 1 dplyr   1.0.8   cran  pkg:cran/dplyr@… "dplyr:  A… https://… <list [0]>                     0

R/assess_security.R
emilliman5, line 1: I think the assessment should be the tibble or possibly the list of vulnerabilities found. and then metric can be binary or presence/absence of vulnerabilities.

R/assess_security.R
```diff
@@ -0,0 +1,55 @@
...
+
+#' Score a package for known security vulnerabilities in the OSS Index
+#'
+#' Coerce the count of reported vulnerabilities to a binary indicator.
+#'
+#' @eval roxygen_score_family("security", dontrun = TRUE)
+#' @return \code{NA} if no vulnerabilities are found, otherwise \code{0}
+#'
+#' @export
+metric_score.pkg_metric_security <- function(x, ...) {
```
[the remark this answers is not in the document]
emilliman5: If we can capture the vulnerability overivew then we could set metric to `NA` if the package is not found in the Sonatype database.

emilliman5 (changes requested):

PR #274, opened 2023-02-10
Add examples available metric

### LINKED ISSUE #43: Examples available for exported objects
Similar to #43, provide an assessment of the examples that are provided with each function.

We could check for the existence of examples or usage fields in the documentation. This can be achieved using the same mechanism of scraping the help as is linked in the references in #43.

parmsam-pfizer:
Fixes #43. Created metric based on our alias finding script.


R/assess_has_examples.R
```diff
@@ -0,0 +1,36 @@
...
+}
+
+# assign a friendly name for examples column
+attributes(assess_has_examples)$column_name <- "has_examples"
+attributes(assess_has_examples)$label <- "number of discovered function files with examples"
+
+#' @export
+assess_has_examples.pkg_ref <- function(x, ...) {
+  pkg_metric_eval(class = "pkg_metric_has_examples", {
+    length(x$examples)
```
elimillera, line 19: Would a better assessment be to find the proportion of objects that have examples rather than the number?

parmsam-pfizer: Good idea. I'll adjust that.

emilliman5:
Nicely done. 

We also need to gracefully handle the condition when a package does not export any objects. You are probably "but Eric why in the world would someone create a package if it does not export anything to the user?" One answer: C headers files. A good example of this is [BH](https://cran.r-project.org/package=BH). 

IMO the ideal behavior if nothing is exported: pkg_metric.assessment should return NA and the assessment should also return NA or a message stating nothing is exported.

emilliman5 (approved):

emilliman5:
@parmsam-pfizer i think you need to resolve the man/pkg_assess.Rd conflict for the checks to run

PR #275, opened 2023-02-22
Prep for 0.2.0 release to cran

elimillera:
Here are some changes to the release. I added some changes to the testing args. The initial tests should fail with the warnings I was seeing.

I updated to version 0.2.0 because we have new functionality and per semantic versioning that warrants a minor number bump.

Also updated the NEWS file.


emilliman5 (approved):
looks good

Issue #282, opened 2023-03-15
assess_downloads_1yr() error


RozetaSimonovska:
assess_downloads_1yr() produces the following:
<simpleError in unclass(as.Date(e1)) - e2: non-numeric argument to binary operator>

Example:
assess_downloads_1yr(pkg_ref("riskmetric"))

riskmetric version 0.2.1
R version 4.2.2 

With riskmetric 0.1.2:
> assess_downloads_1yr(pkg_ref("riskmetric"))
[1] 3368

AbhaySThakur:
I am facing the same issue with the last updates.  Reprex for the same is as below
``` r
library("riskmetric")
assess_downloads_1yr(pkg_ref("riskmetric"))
#> <simpleError in unclass(as.Date(e1)) - e2: non-numeric argument to binary operator>

sessionInfo()
#> R version 4.2.0 (2022-04-22 ucrt)
#> Platform: x86_64-w64-mingw32/x64 (64-bit)
#> Running under: Windows 10 x64 (build 19044)
#> 
#> Matrix products: default
#> 
#> locale:
#> [1] LC_COLLATE=English_United States.utf8 
#> [2] LC_CTYPE=English_United States.utf8   
#> [3] LC_MONETARY=English_United States.utf8
#> [4] LC_NUMERIC=C                          
#> [5] LC_TIME=English_United States.utf8    
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> other attached packages:
#> [1] riskmetric_0.2.1
#> 
#> loaded via a namespace (and not attached):
#>  [1] xfun_0.36           remotes_2.4.2       purrr_1.0.0        
#>  [4] vctrs_0.5.1         miniUI_0.1.1.1      htmltools_0.5.4    
#>  [7] usethis_2.1.6       yaml_2.3.6          utf8_1.2.2         
#> [10] rlang_1.0.6         pkgbuild_1.4.0      later_1.3.0        
#> [13] urlchecker_1.0.1    pillar_1.8.1        glue_1.6.2         
#> [16] withr_2.5.0         sessioninfo_1.2.2   lifecycle_1.0.3    
#> [19] stringr_1.5.0       devtools_2.4.5      htmlwidgets_1.6.0  
#> [22] memoise_2.0.1       evaluate_0.19       knitr_1.41         
#> [25] callr_3.7.3         fastmap_1.1.0       httpuv_1.6.7       
[...]

PR #283, opened 2023-03-22
Fix downloads error

parmsam-pfizer:


emilliman5 (changes requested):
Lets add a warning as well. to let users know there was a problem and that we automatically set the look back

parmsam-pfizer:
Good idea. I'll add that. However, I'm having problems changing the lookback argument. Doesn't look like that is working.

parmsam-pfizer:
Based on our discussion, I've updated `pkg_ref_cache.downloads()` to grab the earliest downloads record date which seems to be ["2012-10-01"](https://github.com/r-hub/cranlogs.app). I wasn't able to find an easy way to grab the package CRAN release date, so used this as an alternative. The `assess_downloads_1yr.pkg_ref()` subsets the pkg_ref downloads dataframe up to the past year then returns the downloads sum.

emilliman5 (changes requested):
Let's add a filter if the "archive_release_dates" field is populated/available. we do not need to display/store download information for dates before a package's first release. This slot is only populated for `pkg_cran_remote` class pkg_refs.

emilliman5:
Almost there.  I think `pkg_ref_cache.downloads.default` should return `NA`. For `pkg_ref_cache.downloads.pkg_cran_remote` I would add some logic to that sets a default from date in case there is no date in `archive_release_dates` (e.g. if there is/was a parsing error). 

and can we implement a downloads ref cache for bioconductor and github???? (can be a later PR)

parmsam-pfizer:
Yeah, I'll create an issue to implement a downloads ref cache for bioconductor and github. I've tried setting 
 pkg_ref_cache.downloads.default to return NA, but the challenge is that is leads to a NA return unless the source is explicitly declared as pkg_cran_remote. Is that the intention with the default class?
pr1 <- pkg_ref(c("dplyr")) %>% pkg_assess() %>% pkg_score()
pr1$downloads

Code snippet I've updated based on your suggestions is below:
```
#' @importFrom cranlogs cran_downloads
#' @keywords internal
pkg_ref_cache.downloads.default <- function(x, ...) {
  return(NA)
}

#' @importFrom cranlogs cran_downloads
#' @keywords internal
pkg_ref_cache.downloads.pkg_cran_remote <- function(x, name, ...) {
  from_date <- tryCatch({
    x$archive_release_dates[1, 'date'][[1]]
    },
    error = function(e){
      # default to downloads from_date will be Oct2012
      # Oct2012 is when RStudio CRAN mirror started publishing logs
    as.Date("2012-10-01", format = "%Y-%m-%d")
    }
  )

  cran_downloads(x$name, from = from_date, to = Sys.Date())
}
```

emilliman5:
I see what you mean.

Maybe there should be no `default` method dispatch??? My thinking is, e.g.,  that for a `pkg_ref_source` we should not even be trying to populate `pkg_ref_cache.downloads` because the `pkg_ref` is not from CRAN. So maybe it is better to have no `pkg_ref_cache.downloads.default` and instead just the `cran_remote` method. and then we add a `github` and `bioc_remote` at a later date? I will have to check what happens when you run an assessment if there is no method to cache the information though.

My goal is to be intentional/explicit about what info we cache based on the ref source. Maybe we need a similar system like for metrics where we specify the reason for the NA???? 

I like the error handling. I need to think more on the method dispatching

parmsam-pfizer:
I'll create a new PR with a temp fix based on [@elimillera's suggestion](https://github.com/pharmaR/riskmetric/issues/282#issuecomment-1511602085). That will at least get the metric working again. Keeping this PR active for us to discuss further, since we're still evaluating the general approach for the metric.

emilliman5:
I will resolve these conflicts and merge

PR #305, opened 2023-07-24
Fix #276 - Create pkg_risk wrapper func

### LINKED ISSUE #276: Consider adding a wrapper function
The wrapper function would cover the three main steps of the workflow below (maybe called `pkg_risk()`). I imagine that most users are just using this workflow, so it makes sense to me to provide a wrapper function. The underlying functions would still be available for users who need to drill in further. Creating issue to continue discussion from today's sprint meeting.
```
pkg_ref(c("riskmetric", "utils", "tools")) %>%
  pkg_assess() %>%
  pkg_score()
```

parmsam-pfizer:
Based on issue #276.


parmsam-pfizer:
@emilliman5, @elimillera 
Please let me know if there are any additional checks, messages, or arguments we want in the wrapper function. Right now, it's calling the core function pattern: `pkg_ref()` to tibble coercion to `pkg_assess()` to `pkg_score()`. There's a simple check for a character vector of package names (the primary arg). Also, there's a message when the default pkg_install source is used.

emilliman5:
Messaging and parameters are a good improvement. I think the default source should be "pkg_cran_remote" this will reduce the number of errors when someone supplies a long-ish vector of packages where some are installed and some not. Longer term, we will need some logic (messaging or heuristics) to handle the case of different sources. A few options:

1) Error and force user to only provide packages that are available from same source (not nescessarily the most user friendly)
2) remove packages that are not available from given source and provide message
3) instaniate pkg_ref for packages with preferred source (based on argument value). for any packages not available fall back to another source
    e.g. `pkg_risk(c("dplyr","spongebob"))` `dplyr` is installed so use "pkg_install" but `spongebob` is not so try CRAN, then Bioc then github.
4) something even better that i have not thought of

emilliman5 (changes requested):
I get an error if the package is not already installed. If packages are not installed then it should try cran_remote, bioc_remote, etc. There is an old `pkg_ref()` function that had a simple heuristic to build the pkg_ref object based on installed, remote, source, etc. It might be good to use that to initial the pkg_ref but with messaging

parmsam-pfizer:
Thanks for the suggestions, @emilliman5. Updated branch to reflect our latest discussion in this PR. Updated documentation and created `pkg_risk_cran()` as an alternative for folks who want to ensure they point to "pkg_cran_remote". Hopefully, this will also inform users about the difference source types. 

I like the nature of the pkg_cran_remote class that retains a row for the requested package that aren't found with a pkg_missing response in the pkg_ref column. Example of this: 
`pkg_risk_cran(c("riskmetric", "dplyr", "abc", "doesntexist"))`


I'd like to replicate that for the pkg_install source type. Right now, it errors out when the "pkg_install" source is used but the package isn't found. What are your thoughts?
`pkg_risk(c("riskmetric", "dplyr", "doesntexist"), source = "pkg_install")`

parmsam-pfizer:
Hi, @emilliman5! I'd love to know your thoughts on the changes when you get a chance. Thanks!

emilliman5:
I really like `pkg_risk_cran`! A few thoughts:

1) Can we make one function to handle all sources (like this `pkg_risk(c("riskmetric", "dplyr", "doesntexist"), source = "pkg_install")`)
2) Doug had written a simple heuristic to handle the error you describe above in an early version of `pkg_ref` [I think may be a good starting point](https://github.com/pharmaR/riskmetric/blob/d47eae35778aae1d19dba490196ad7a6cfa7c7a1/R/pkg_ref_class.R#L125)

Lastly, My apologies for radio silence. I think I have finally freed up some bandwidth to dedicate to riskmetric once again.

parmsam-pfizer:
Sorry for the delay. Made a few updates to handle the error when the "pkg_install" source is used but the package isn't found. It should now retain a row for the requested package that wasn't found. It does this via a check on `installed.packages()` before running `pkg_ref()`. For 3, I added a simple function called `fix_src_name()` to auto adjust the source type. This should enable users to pass just "cran" or "install" and get the right source type.

emilliman5 (approved):

parmsam:
Merging to `ease_of_use` branch. Assuming this is safe since it is not yet in main and PR is approved. Thanks!

PR #353, opened 2024-09-17
updated has_examples documentation to better reflect what is being as…

emilliman5:
…sessed and scored


parmsam-pfizer (commented):
Hi @emilliman5. Looks good to me. Should the roxygen2 header for `assess_has_examples()` also be updated? See line 5.

Issue #360, opened 2024-11-28
Bioconductor package downloads


AKlebus:
Feature idea: add functionality to get package download metrics from Bioconductor packages. Currently this only works for CRAN packages.

llrs-roche:
This could be relatively easy with the [BiocPkgTools](https://www.bioconductor.org/packages/BiocPkgTools) package from Bioconductor (or picking how they do it if riskmetrics doesn't want to add it as a dependency). HTH

emilliman5:
I was unaware the BioC was collecting download metrics. If you create a PR I will happily review and merge. This appears to be a simple add.

PR #361, opened 2024-12-02
Bioconductor downloads

AKlebus:
Issue: #360 

# Bioconductor downloads
This PR adds functionality for caching downloads for packages on Bioconductor. Previously, this downloads functionality only attempted to get package downloads from CRAN, meaning popular Bioconductor packages were judged as if they had 0 downloads. With this change, assessments of Bioconductor package downloads should be more informative.

## Background
Hi! My team and I at Novartis wanted to contribute this functionality to riskmetric as it is something we have implemented for an internal tool. I personally am quite new to software engineering, and this is my first PR for an open source project, so I would really appreciate any thoughts or feedback on how we can better integrate this functionality. Looking forward to the discussion!

### DIFF
--- a/R/pkg_ref_cache_downloads.R
+++ b/R/pkg_ref_cache_downloads.R
@@ -6,29 +6,29 @@
 #' @return a \code{pkg_ref} object
 #' @keywords internal
 #' @noRd
-pkg_ref_cache.downloads <- function(x, ..., n=365) {
+pkg_ref_cache.downloads <- function(x, ..., n = 365) {
   UseMethod("pkg_ref_cache.downloads")
 }
 
 #' @importFrom cranlogs cran_downloads
 #' @keywords internal
-pkg_ref_cache.downloads.pkg_ref <- function(x, ..., n=365) {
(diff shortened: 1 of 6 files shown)


R/pkg_ref_cache_downloads.R
```diff
@@ -1,13 +1,97 @@
...
-pkg_ref_cache.downloads <- function(x, ..., n=365) {
-  cran_downloads(x$name, from=Sys.Date()-n, to=Sys.Date())
+pkg_ref_cache.downloads <- function(x, ..., n = 365) {
+  UseMethod("pkg_ref_cache.downloads")
+}
+
+#' @importFrom cranlogs cran_downloads
+#' @keywords internal
+pkg_ref_cache.downloads.pkg_ref <- function(x, ..., n = 365) {
+  downloads <- cran_downloads(x$name, from = Sys.Date() - n, to = Sys.Date())
```
emilliman5, line 16: I don't think there is any need to dispatch a download method for `pkg_ref`, in fact we should return NA "by default" to account for sources that do not have download stats (e.g. source packages)

emilliman5 (changes requested):

PR #367, opened 2025-03-06
Updates to resolve CRAN comments in R-devel

elimillera:
From [R-devel news](https://cran.r-project.org/doc/manuals/r-devel/NEWS.html):

> grep(), strsplit() and similar took non-TRUE values of their logical arguments as FALSE, but these were almost always mistakes and are now reported as NA.

There was an error in the dependency check for source packages. and removing the extra `strsplit` fixes it

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -34,7 +34,6 @@ jobs:
           - {os: ubuntu-22.04,   r: 'oldrel-1'}
           - {os: ubuntu-22.04,   r: 'oldrel-2'}
           - {os: ubuntu-22.04,   r: 'oldrel-3'}
-          - {os: ubuntu-22.04,   r: 'oldrel-4'}
           - {os: ubuntu-20.04,   r: 'release'}
 
     env:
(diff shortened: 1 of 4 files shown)


dgkf (approved):
Thanks for jumping on this @elimillera! LGTM

And just in the way of sharing a maintenance update, we're trying to be more mindful of having a more consistent and intentional maintenance model across all the R Val Hub efforts. One way we're trying to align packages is to use a maintainer mailing list under the R Consortium. Would that be okay with you?

Still a bit of uncertainty - I'm not sure how CRAN feels about a mailing list as a maintainer email - so I don't think a time-sensitive release is the right place to make that jump. 

Regardless let me know if you would be okay with that change. To eventually hand over maintainers we would just need you to confirm during a version release. Even then, we'd be happy to have you joining that mailing list if you're interested so that it would just be a matter of you CC'ing yourself on a package submission under a different email address.

elimillera:
Hey @dgkf, That is fine for me in theory if you can get CRAN to agree to it. From the [CRAN policies](https://cran.r-project.org/web/packages/policies.html) they do say this is not allowed:

> The package’s DESCRIPTION file must show both the name and email address of a single designated maintainer (a person, not a mailing list). That contact address must be kept up to date, and be usable for information mailed by the CRAN team without any form of filtering, confirmation …. Forwarding mail from the maintainer address increasingly results in confusing non-delivery notifications to the original sender, so is best avoided

So happy to be a part of any maintainer list but I'm doubtful that CRAN will allow that

PR #368, opened 2025-03-13
Remove package_version

### LINKED ISSUE #364: `pkg_ref` alters semantic versioning of source packages
Some packages on CRAN follow `x.y-z` semantic versioning convention, for example [sf](https://cran.r-project.org/web/packages/sf/index.html), [abind](https://cran.r-project.org/web/packages/abind/).

`riskmetric::pkg_ref` changes the semantics to be `x.y.z` due to the usage of `base::package_version()` specifically [when working with the source packages](https://github.com/pharmaR/riskmetric/blob/39489b26dbb3b7e5731bbcf55d975755fbb841e4/R/pkg_ref_class.R#L167).

I think it is conceptually not ideal, because later it is not possible to join assessment data with the original package DESCRIPTION meta info.

I am opening this issue to understand better why the line of code linked above is necessary, and why can't we use a string representation of the version. I am definitely missing some broader context here.

My initial thought was version comparisons, but it works fine with the standard library function:

```r
version_old <- "1.1-1"
version_new <- "1.1-2"
utils::compareVersion(version_new, version_old)
```

cc @llrs-roche

llrs-roche:
Removes the `package_version` casting. 

This will enable it to work on old versions of packages before there was a requirement on what was allowed as version number. 
Note: R doesn't force semantic versioning (three numbers). 

Closes #364

### DIFF
--- a/R/pkg_ref_cache_archive_release_date.R
+++ b/R/pkg_ref_cache_archive_release_date.R
@@ -18,8 +18,7 @@ pkg_ref_cache.archive_release_dates.pkg_cran_remote <- function(x, name, ...) {
 
   text <- unlist(strsplit(xml2::xml_text(node), "\n"))
   db   <- do.call(rbind, strsplit(text[-1], "\\s+"))
-  version <- package_version(gsub(paste0(x$name, "_(.*)\\.tar\\.gz"), "\\1", db[,2]))
-  version <- as.character(version)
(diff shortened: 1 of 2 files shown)


dgkf (approved):
looks great. Many thanks to you and @Gotfrid for teasing out this problem and helping to improve the tool for everyone.

PR #369, opened 2025-03-13
Check that suggested packages used on tests are available

llrs-roche:
Many suggested packages are used unconditionally on tests:
magrittr pipe is used  in tests/testthat/setup_mock_web_requests.R
withr in tests/testthat/setup_test_packages.R
webmockr is used  in tests/testthat/setup_mock_web_requests.R
jsonlite is used  in tests/testthat/setup_mock_web_requests.R

This PR checks they are available before running any test

### DIFF
--- a/tests/testthat.R
+++ b/tests/testthat.R
@@ -1,6 +1,8 @@
-library(testthat)
-library(riskmetric)
+if (requireNamespace("testthat", quietly = TRUE) && requireNamespace("webmockr", quietly = TRUE) && requireNamespace("jsonlite", quietly = TRUE) && requireNamespace("magrittr", quietly = TRUE)  && requireNamespace("withr", quietly = TRUE)) {
+  library(testthat)
+  library(riskmetric)
 
-options(repos = "fake-cran.fake-r-project.org")
+  options(repos = "fake-cran.fake-r-project.org")
 
-test_check("riskmetric")
+  test_check("riskmetric")
+}


dgkf:
I'm not so sure that this is a problem. The [R extensions manual](https://cran.r-project.org/doc/manuals/R-exts.html#Package-Dependencies-1) says

> The `Suggests` field uses the same syntax as ‘Depends’ and lists packages that are not necessarily needed. This includes packages used only in examples, tests or vignettes

That said, it can be frustrating to run tests and incrementally discover that you're missing dependencies. To address that developer-facing pain point, I would instead throw an error that lists necessary testing packages. I'm going to commit some changes to implement this behavior, but if you feel strongly against my direction we can revert and find a different path.

dgkf:
Reworked the PR, I hope you don't mind. I wanted to enhance this with a few new design goals:

1. Ideally we check all the dependencies upfront so that the failure mode reports all missing packages, preventing a developer experience that requires iteratively running tests and discovering new missing dependencies.
2. Run as part of `testthat`, so that we perform the check whether we use `tools::testInstalledPackage` (which runs `tests/*.R`) as well as via `testthat::test_package()` via something like `devtools`.
3. Instead having to remember to update this list somewhere buried in the testing files, leverage the [`Config/Needs` convention](https://pak.r-lib.org/reference/package-dependency-types.html?q=Config%20Needs#extra-dependencies) to document dependencies for a specific purpose.

Let me know what you think.

llrs-roche:
While this is good and works, the specific section about suggested packages on [WRE](https://cran.r-project.org/doc/manuals/R-exts.html#Suggested-packages-1) says something differently:

> The recommendation used to be to make their use conditional...
> It is better practice to use code like
> 
> ```r
>    if (requireNamespace("rgl", quietly = TRUE)) {
>       rgl::plot3d(...)
>    } else {
>       ## do something else not involving rgl.
>    }
> ```

In addition it requires additional tools (pak) to resolve the situation.   

> Reworked the PR, I hope you don't mind. I wanted to enhance this with a few new design goals:
> 
> 1. Ideally we check all the dependencies upfront so that the failure mode reports all missing packages, preventing a developer experience that requires iteratively running tests and discovering new missing dependencies.

If we want to ensure that all packages are upfront move them to `'Imports'`. It won't be possible to install or check the package without them (and won't require an additional tool). 

> 2. Run as part of `testthat`, so that we perform the check whether we use `tools::testInstalledPackage` (which runs `tests/*.R`) as well as via `testthat::test_package()` via something like `devtools`.

Adding the check on a specific tests/testthat/*.R files already requires testthat to be available. If testthat is not present what will happen? (This happened a while back on CRAN). `tools::testInstalledPackage` works if tests are installed, from the help pages:

> If package-specific tests are found in a ‘tests’ directory they can be tested: these are not installed by default, but will be if R CMD INSTALL --install-tests was used

The original approach work for `R CMD check` and any other method. 

> 3. Instead having to remember to update this list somewhere buried in the testing files, leverage the [`Config/Needs` convention](https://pak.r-lib.org/reference/package-dependency-types.html?q=Config%20Needs#extra-dependencies) to document dependencies for a specific purpose.
> 
> If any package is missing, or if its version constraints aren't met, testthat will fail early with an error message that looks like:
> 
> ```
> Error:
>   Missing Suggests packages required for testing:
>     testthat (<= 1.2.3)
> 
>   Resolve using:
>     pak::local_install_dev_deps(dependencies = "Config/Needs/testing")
> ```
> 
> Let me know what you think.

3. I think adding an additional field would make sense for packages that are not on the package but are used to build some parts of it (roxygen2 macros, ...) or for additional checks on the CI (styler, covr, ...) . Currently it duplicates the packages from Suggests and I'm not sure it reduces the maintenance efforts.

dgkf:
Ah, thanks for the tips! Some great points in here.

1. You're right `devtools::test` _does_ pick up `tests/testthat.R`. I had misremembered, it _does_ run `tests/testthat.R`, but it _does not_ run other files in `tests`.
2. And thanks for the correction on the extensions manual. I thought the defensive programming style was only a recommendation for the main package code. Thanks for pointing out the specific recommendation for tests as well.
3. I really like the `Config/needs` pattern, but can concede it here.

I'll revert the changes, though I'd still prefer a mechanism that checks _all_ required dependencies upfront. I'll see if I can find a more minimalist middle ground.

dgkf:
After doing more investigation, I think I'm more convinced that my approach is necessary if we want to check all suggested dependencies upfront.

Alternatively, we could add `testthat::skip_if_not_installed(...)` to gate execution for tests that require specific suggests dependencies.

On specific design choices, 

- we can drop the `Config/needs` in lieu of listing packages within the tests
- we can drop the suggestion to use `pak` in lieu of `install.packages(...)`

Pushed a change that simplifies it more in line with your initial approach.

llrs-roche:
I think `skip_if_not_installed` is a good compromise, it provides automatic messages when failing to find the pacakges. 

> Okay, I think I was right the first time around. `devtools::test` does not run `tests/testthat.R`, meaning these checks will be skipped. They need to live inside tests/testthat/ since `devtools::test` internally runs `testthat::test_local()` which only runs files within tests/testthat/.

Oh, didn't expect that, sorry for the misdirection.

PR #370, opened 2025-03-13
Export methods so that roxygen2 doesn't complain

### LINKED ISSUE #357: Missing exported functions pkg_*
I was reading `?pkg_ref` and realized that there are several functions documented but not exported (and only can be used with `riskmetric:::pkg_*`:

```r
pkg_install(x, lib.loc = NULL)
pkg_source(x)
pkg_cran(x, repos = getOption("repos", "https://cran.rstudio.com"))
pkg_bioc(x)
pkg_missing(x)
pkg_library(lib.loc)
```

Maybe they could be exported or avoid documenting them as it is a bit confusing for a new user. I can put together a PR to export them.

llrs-roche:
When updating and using the latest roxygen2 version I noticed that there are several warnings regarding methods and exporting.
This PR fixes them, it also fixes some notes related to the documentation when checking the package (\describe, keywords)  but it doesn't add documentation for functions:

This would close #357
```r
Undocumented code objects:
  'bug_reports_status' 'metric_score_condition'
  'pkg_ref_cache.bug_reports' 'pkg_ref_cache.examples'
  'pkg_ref_cache.help_aliases' 'pkg_ref_cache.license'
  'pkg_ref_cache.news' 'pkg_ref_cache.repo_base_url'
  'pkg_ref_cache.tarball_url' 'pkg_ref_cache.vignettes'
  'pkg_ref_cache.web_url' 'pkg_ref_cache.website_urls'
  'scrape_bug_reports'
```

Let me know if you want me to document those methods/functions


dgkf (approved):
Thanks for tackling this! And today I learned that you can use `@export` + `@method`! I've always used `@exportS3Method`, but it feels so clunky that way.

dgkf:
@llrs-roche - I went through and resolved remaining issues. Not entirely sure what the best practices for double dispatch documentation are, so I'd appreciate your review on my changes.

R/pkg_score.R
```diff
@@ -112,7 +109,7 @@ roxygen_score_family <- function(name, dontrun = TRUE) {
       "package. Please provide one to complete documentation."),
       name, assess_func))
 
-  if (!score_func %in% getNamespaceExports(utils::packageName()))
+  if (!score_func %in% .S3methods("metric_score"))
```
llrs-roche, line 112: Nice, wasn't sure how to fix this.

llrs-roche (commented):
Looks good, you extended the PR to fix all the documentation issues (I had a different branch prepared for that). Thanks!!

I left one minor comment. I'll check this more carefully later this week, as I'm not sure what happens with internal documentation with no usage (and I want to check the patterns here because I think some have noRd and others do have documentation file created).

R/pkg_ref_cache_downloads.R
```diff
@@ -1,13 +1,12 @@
...
 #'
 #' @importFrom cranlogs cran_downloads
-#' @return a \code{pkg_ref} object
-#' @keywords internal
-#' @noRd
-pkg_ref_cache.downloads <- function(x, ..., n=365) {
-  cran_downloads(x$name, from=Sys.Date()-n, to=Sys.Date())
+#' @usage NULL
+#' @export
+pkg_ref_cache.downloads <- function(x, ..., n = 365) {
```
llrs-roche, line 10: The generic has a name argument:

```suggestion
pkg_ref_cache.downloads <- function(x, name, ..., n = 365) {
```

Not sure if this is why the help page shows up as:

> `pkg_ref_cache(downloads)`: Cache a list of available help files as LaTeX objects

It also happens with `pkg_ref_cache(source_control_url)`. 

On that same page (`?pkg_ref_cache`),  there are many lines repeated on the See Also: "Other package reference cache"

R/pkg_ref_cache_NEWS_urls.R
```diff
@@ -1,17 +1,19 @@
...
 #' Cache appropriate urls for NEWS files
 #'
-#' @inheritParams pkg_ref_cache
 #' @family package reference cache
-#' @return a \code{pkg_ref} object
 #' @keywords internal
-#' @noRd
+#'
+#' @usage NULL
+#' @export
```
llrs-roche, line 8: If it is internal and documented on another help page I think it only needs to have

```
#' @describeIn riskmetric_metadata_caching
#' @export
```

It is not really internal and other documentation (should)/is already on the `riskmetric_metadata_caching` help page. This will probably remove the family tag (besides it is a small family of just one other function, so a simple link to it would be enough).

llrs-roche (commented):
I don't find the documentation you refer to as double dispatch. I think with S3 it is not possible to do it.  `NextMethod()` only allows one object to define which method is used. 
The closest one I can find is with operators (?groupGeneric ), but I don't see any group or operators defined on this package. 

Besides that I made a couple more comments

dgkf:
> I don't find the documentation you refer to as double dispatch. I think with S3 it is not possible to do it. `NextMethod()` only allows one object to define which method is used. The closest one I can find is with operators (?groupGeneric ), but I don't see any group or operators defined on this package.
> 
> Besides that I made a couple more comments

Ah, sorry - yes you're right. S3 doesn't support it out of the box. It's a pattern of using S3 where you dispatch on one parameter (`add.my_class`), but then that dispatches again on another argument (`add.my_class.your_class`). To my knowledge, it was popularized in `vctrs` (see [this vignette](https://vctrs.r-lib.org/articles/s3-vector.html?q=dispatch#double-dispatch)).

llrs-roche (commented):
I've read the vignette and checked the implementation here. As you can see I'm not familiar with that approach, consider this when reading my comments below.
 
This documentation approach for the "double dispatch" methods is only needed for certain vectors methods (arithmetics, see https://vctrs.r-lib.org/articles/s3-vector.html?q=dispatch#arithmetic-1) so I'm not sure if documenting via `@method` is expected here or not. 

In any case, using `#' @method pkg_ref_cache downloads` on `pkg_ref_cache.downloads` improved the documentation for it. But applying the same for `pkg_ref_cache.source_control_url` did not. It might be related to  missing vec_type2 method definition:

"Once you’ve implemented [vec_ptype2()](https://vctrs.r-lib.org/reference/vec_ptype2.html) and [vec_cast()](https://vctrs.r-lib.org/reference/vec_cast.html), you ..."
Which should be used to convert from one of the classes to the other (?).
pkg_ref_cache might need a `vec_cast` and `vec_ptype2` methods to work well (It currently  misses both of them). 

Either way the checks on the documentation pass fine. However, I don't understand the benefits of using this approach over simpler ones.

emilliman5 (approved):

PR #371, opened 2025-03-14
Vignette improvements

llrs-roche:
This reduces calculation of some steps on the vignette.
Locally it was failing because `has_news` was not available. I changed to a different metric. 
I think `pkg_score` would benefit from some defensive programming (checking that the tibble has more than 1 column), but I haven't pushed that change.


llrs-roche:
I mentioned the defensive programming but it is not included on this PR. This PR only avoid recalculating the assessment multiple times to make it quicker to build and check.

emilliman5 (approved):

PR #382, opened 2025-04-23
bugfix: assess_export_help.pkg_source

### LINKED ISSUE #381: assess_export_help.pkg_source() fails when dependencies of target package are not installed
The function `assess_export_help.pkg_source()` fails when the target package (i.e., the one being assessed) has dependencies that are not installed. This is due to the use of `pkgload::parse_ns_file()`, which attempts to evaluate `import()` and other namespace directives, thus requiring all dependencies to be present.

This makes the metric fragile and unsuitable for use in minimal environments (e.g., CI pipelines or isolated testing setups) where not all dependencies of a package are guaranteed to be installed.

### Minimal example

When assessing the package `ggh4x`, which imports `ggplot2`, and `ggplot2` is not installed, the function returns the following error structure instead of the expected logical vector:

```
str(output)
List of 4
 $ message: chr "there is no package called ‘ggplot2’"
 $ call   : language packageVersion("ggplot2")
 $ package: chr "ggplot2"
 $ lib.loc: NULL
 - attr(*, "class")= chr [1:8] "pkg_metric_error" "pkg_metric_condition" "pkg_metric_export_help" "pkg_metric" ...
 - attr(*, ".recording")=List of 5
  ..$ expr      : language {  exports <- unname(unlist(pkgload::parse_ns_file(x$path)[c("exports",; "exportMethods")])); out <- exports %in%| __truncated__
  ..$ attributes:List of 2
  .. ..$ names: chr [1:4] "message" "call" "package" "lib.loc"
  .. ..$ class: chr [1:3] "packageNotFoundError" "error" "condition"
  ..$ visible   : logi TRUE
  ..$ traceback :List of 8
  .. ..$ : language packageVersion("ggplot2")
  .. ..$ : language eval(e[[2L]], .GlobalEnv)
  .. ..$ : language parseDirective(e)
  .. ..$ : language parseNamespaceFile(basename(path), dirname(path), mustExist = FALSE)
  .. ..$ : language pkgload::parse_ns_file(x$path)
  .. ..$ : language unlist(pkgload::parse_ns_file(x$path)[c("exports", "exportMethods")])
  .. ..$ : language unname(unlist(pkgload::parse_ns_file(x$path)[c("exports", "exportMethods")]))
  ..$ output    :List of 1
  .. ..$ :List of 4
  .. .. ..$ message: chr "there is no package called ‘ggplot2’"
  .. .. ..$ call   : language eval(e[[2L]], .GlobalEnv)
  .. .. ..$ package: chr "ggplot2"
  .. .. ..$ lib.loc: NULL
  .. .. ..- attr(*, "class")= chr [1:3] "packageNotFoundError" "error" "condition"
 - attr(*, "label")= chr "exported objects have documentation"
```

When `ggplot2` is installed, the function works as expected:

```
str(output)
 'pkg_metric_export_help' Named logi [1:83] TRUE TRUE TRUE TRUE TRUE TRUE ...
 - attr(*, "names")= chr [1:83] "CoordAxesInside" "FacetGrid2" "FacetManual" "FacetNested" ...
 - attr(*, ".recording")=List of 5
  ..$ expr      : language {  exports <- unname(unlist(pkgload::parse_ns_file(x$path)[c("exports",; "exportMethods")])); out <- exports %in%| __truncated__
  ..$ attributes:List of 1
  .. ..$ names: chr [1:83] "CoordAxesInside" "FacetGrid2" "FacetManual" "FacetNested" ...
  ..$ visible   : logi TRUE
  ..$ traceback : NULL
  ..$ output    : chr(0)
 - attr(*, "label")= chr "exported objects have documentation"
```

rabiibouhestine:
close #381 

The function `assess_export_help.pkg_source()` fails when the target package (i.e., the one being assessed) has dependencies that are not installed. This is due to the use of `pkgload::parse_ns_file()`, which attempts to evaluate `import()` and other namespace directives, thus requiring all dependencies to be present.

This makes the metric fragile and unsuitable for use in minimal environments (e.g., CI pipelines or isolated testing setups) where not all dependencies of a package are guaranteed to be installed.

### DIFF
--- a/R/assess_export_help.R
+++ b/R/assess_export_help.R
@@ -28,7 +28,7 @@ assess_export_help.pkg_source <- function(x, ...) {
   pkg_metric_eval(class = "pkg_metric_export_help", {
     # ignore S3-dispatched methods
     lines <- readLines(paste0(x$path, "/NAMESPACE"), warn = FALSE)
-    export_lines <- grep("^export\\(|^exportMethods\\(", lines, value = TRUE)
+    export_lines <- grep("^export\\(|^exportMethod\\(", lines, value = TRUE)
     exports <- gsub(".*\\(([^)]+)\\).*", "\\1", export_lines)
     exports <- unlist(strsplit(exports, ",\\s*"))
 
--- a/R/assess_export_help.R
+++ b/R/assess_export_help.R
@@ -27,7 +27,11 @@ assess_export_help.pkg_remote <- function(x, ...) {
 assess_export_help.pkg_source <- function(x, ...) {
   pkg_metric_eval(class = "pkg_metric_export_help", {
     # ignore S3-dispatched methods
-    exports <- unname(unlist(pkgload::parse_ns_file(x$path)[c("exports","exportMethods")]))
+    lines <- readLines(paste0(x$path, "/NAMESPACE"), warn = FALSE)
+    export_lines <- grep("^export\\(|^exportMethods\\(", lines, value = TRUE)
+    exports <- gsub(".*\\(([^)]+)\\).*", "\\1", export_lines)
+    exports <- unlist(strsplit(exports, ",\\s*"))
+
     out <- exports %in% names(x$help_aliases)
     names(out) <- exports
     out


llrs-roche (commented):
Nice finding!

Since the extraction is now manual and doesn't use pkgload does it work also for packages using [`exportPattern` on their NAMESPACE](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Specifying-imports-and-exports-1) ? 
I am not sure it worked previously for those packages (If you need to test here are a [list of packages that use this](https://github.com/search?q=org%3Acran%20path%3ANAMESPACE%20exportPattern&type=code)) but if we are modifying this we might be able to fix two things at once. 

I think it makes more sense to ensure that the dependencies are installed, as the local check assessment requires at least the strong dependencies. But maybe other people has some other feedback/opinion (@dgkf, @emilliman5 ) ?

llrs-roche:
> @llrs-roche the extraction using `pkgload` only looks at `export` and `exportMethod` so I kept it that way. My change currently does not work with `exportPattern`, but I guess we can also manually extract patterns as well. 

Yes, better to preserve the logic at least now. But I think this is a separate bug that maybe we need to record (or double check it is not already reported) and fix somewhere else. 

I haven't tested this works,  could you add a test to verify this fixes the problem? Probably modifying one of the package on `tests/testthat/test_packages/`  or adding a new one.

> Now that you've mentioned `exportPattern`, Is there a reason for other exports to not be included? like `exportS3Method` and `exportClass`?

Methods are specific for classes and the generic is enough for the general purpose (I think this is already covered by `export` and/or `exportMethod`). I have more doubts with `exportClass` as knowing if a package provides a class could be important (in case it is renamed of packages are split or classes disappear). But this might be a different problem/bug

llrs-roche (approved):
I verified this works well for ggh4x:

```
x <- pkg_source("~/../Downloads/ggh4x_0.3.0/ggh4x/")
a <- pkg_assess(x, assessments = all_assessments()["assess_export_help"])
table(a$export_help)
## TRUE
##   83
``` 

Please open a bug report for `exportedPattern` and discussing the difference between namespace and alias (the relationship can be n to m). But if you wish I'll open it myself.  

And it doesn't break on package with exportPattern (picked NMF):

```
x <- pkg_ref("NMF")
a <- pkg_assess(x, assessments = all_assessments()["assess_export_help"])
table(a$export_help)
## FALSE  TRUE 
##   113   174 
```

PR #384, opened 2025-04-23
remove deprecated ubuntu 20.04 from gha

### LINKED ISSUE #383: Update github actions image for Ubuntu
There is a message to update GHA being used on this repository:

> This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101

rabiibouhestine:
close #383 

Ubuntu 20.04 LTS actions runner image is now [deprecated](https://github.com/actions/runner-images/issues/11101).

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -34,7 +34,6 @@ jobs:
           - {os: ubuntu-22.04,   r: 'oldrel-1'}
           - {os: ubuntu-22.04,   r: 'oldrel-2'}
           - {os: ubuntu-22.04,   r: 'oldrel-3'}
-          - {os: ubuntu-20.04,   r: 'release'}
 
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}


llrs-roche (approved):
Hi, this looks fine to resolve the problem the easiest way.

However, this check seems to come from the [check-full on r-lib/actions](https://github.com/r-lib/actions/blob/v2-branch/examples/check-full.yaml)
There they currently use ubuntu-latest instead of naming the specific version. 

Maybe instead of naming the version we can use the same strategy as the current version: keep "os: ubuntu-latest". Not sure if it is best to update the whole action to avoid the same problem in a couple of years.

rabiibouhestine:
@llrs-roche ubuntu-latest is 24 I believe, I thought there was a need for ubuntu 22.04 to be specifically supported. The comment there says "Use older ubuntu to maximise backward compatibility". But if that's not the case, then I agree it's better to use ubuntu-latest.

llrs-roche:
Yes, ubuntu latest is version 24 but I am not aware of a requirement for a specific OS version. Let's merge as is (If you can't merge it let me know and I'll press the button).

Issue #376, opened 2025-04-04
Conversion to and from PACKAGES file


dgkf:
**Developer day outcome:**  
Allow package data to be able to be round-trip'ed to a `PACKAGES` file

In support of a conventional repository of package qualities, we decided to make the `PACKAGES` file (dcf format, similar to `DESCRIPTION`) the common denominator for sharing package metadata. Writing user-facing data (data which may be used to form decisions) out to a `PACKAGES` file and reading it back in to R should produce the same object that we may use within `riskmetric` itself for decisions.

This decision has a few ramifications:

- Scalars are preferred because it's the least ambiguous data to write and read
- Vectors are likely needed as well. Some investigation needed to standardize a style, but we initially took the dependencies fields as inspiration (eg, `a, b, c` comma separated list)
- We would like error values to be able to pass through this format. Perhaps as error codes or as R code (similar to `Authors@R`)

Even before we have a full implementation of this, I think it would allow other teams to move forward if we could produce a minimal example of what we're aiming for.

dgkf:
Just in terms of syntax, I think metrics should have a prefix of some sort to limit conflict with other fields.

Something like `MetricHasNews` or `RiskmetricHasNews`

dgkf:
Coming from the reporting team, it would be nice to store the download url of the tarball

https://github.com/pharmaR/riskreports/pull/31#discussion_r2042190204

llrs-roche:
Not to derail from this issue/objective, but I found a related file on Bioconductor. It is a JSON file with some of metrics riskmetrics calculates too:

https://bioconductor.org/packages/json/3.20/bioc/packages.json

```
    "hasREADME": false,
    "hasNEWS": true,
    "hasINSTALL": false,
    "hasLICENSE": false,
    ....
    "dependencyCount": "10",
```

Maybe we can find a common ground so that we work with them and the repository could provide some of these metrics (and CRAN in the future). 

I think there is a VIEWS file I cannot locate where they provide the same or similar information in a DCF format, but I haven't found it again.

llrs-roche:
Hi Rabii, thanks for looking into this.
Yes, the DCF file format is mostly a plain text file with `key: value` structure, this is the challenge of this issue. 
I think the JSON format has the same problem with attributes and S3 class metadata (but it handles better the lists).

[...]

PR #386, opened 2025-05-08
Exporting metrics

llrs-roche:
PR to address #376 

Challenges:

1) exporting metrics directly to PACAKGES results on floating numbers with different accuracy (on a test case between 17 and 19 digits)
2) To recover the same object (aka `identical(metrics, metrics |> write() |> read())`) means restoring the labels and some attributes of the metrics:

```
attributes(metrics$downloads_1yr)
## $.recording
## $.recording$expr
## {
##     sum(x$downloads$count)
## }
## 
## $.recording$attributes
## NULL
## 
## $.recording$visible
## [1] TRUE
## 
## ...
```

### DIFF
--- a/R/export.R
+++ b/R/export.R
@@ -1,0 +1,39 @@
+serialize_with_attributes <- function(x) {
+  if (is.list(x) && !is.null(names(x))) {
+    result <- list()
+    for (n in names(x)) {
+      result[[n]] <- serialize_with_attributes(x[[n]])
+    }
+    return(result)
+  }
+
+  # For atomic or S3 elements with attributes
+  val <- unclass(x)
+  attrs <- attributes(x)
+
+  # Sanitize attributes (no S3 objects inside)
+  if (!is.null(attrs)) {
+    for (a in names(attrs)) {
+      if (inherits(attrs[[a]], "AsIs") || is.factor(attrs[[a]])) {
+        attrs[[a]] <- as.character(attrs[[a]])
+      } else if (!is.atomic(attrs[[a]])) {
+        attrs[[a]] <- unclass(attrs[[a]])
+      }
+    }
+  }
+
+  list(
+    value = val,
+    attributes = attrs
+  )
+}
+
+
+export_json <- function(metrics, file = "PACKAGES.json") {
+  metrics_serialized <- serialize_with_attributes(metrics)
+  jsonlite::write_json(metrics_serialized, path = file, pretty = TRUE, auto_unbox = FALSE)
+}
+
+export_dcf <- function(metrics, file = "PACKAGES") {
+  write.dcf(metrics, file = file)
+}


rabiibouhestine:
@dgkf @llrs-roche  Following our discussion last Friday, I wrote a function `pkg_metric_export()` that exports scored package metrics into a flat list of character strings in DCF. It supports both output classes of `pkg_score()` (`list` or `tbl_df`). 

An example output for a single package:

```
testOne <- pkg_ref(utils) %>%
  pkg_assess() %>%
  pkg_score() %>%
  pkg_metric_export()
```

```
MetricPkgScore: 0.6870599
MetricCovrCoverage: NA
MetricHasNews: 0
MetricRemoteChecks: NA
MetricNewsCurrent: 0
MetricRCmdCheck: NA
MetricExportedNamespace: 0.07135333
MetricHasVignettes: 1
MetricExportHelp: 1
MetricHasWebsite: 0
MetricHasMaintainer: 1
MetricBugsStatus: NA
MetricSizeCodebase: NA
MetricHasSourceControl: 0
MetricHasBugReportsUrl: 0
MetricDownloads1yr: 0.00000
MetricReverseDependencies: 1
MetricHasExamples: 0.9937107
MetricDependencies: 0.88079708
MetricLicense: NA
```



An example output for multiple packages:

```
testMany <- pkg_ref(c("utils", "dplyr")) %>%
  pkg_assess() %>%
  pkg_score() %>%
  pkg_metric_export()
```

```
Package: utils
Version: 4.4.3
MetricPkgScore: 0.6870599
MetricCovrCoverage: NA
MetricHasNews: 0
MetricRemoteChecks: NA
MetricNewsCurrent: 0
MetricRCmdCheck: NA
MetricExportedNamespace: 0.07135333
MetricHasVignettes: 1
MetricExportHelp: 1
MetricHasWebsite: 0
MetricHasMaintainer: 1
MetricBugsStatus: NA
MetricSizeCodebase: NA
MetricHasSourceControl: 0
MetricHasBugReportsUrl: 0
MetricDownloads1yr: 0.00000
MetricReverseDependencies: 1
MetricHasExamples: 0.9937107
MetricDependencies: 0.88079708
MetricLicense: NA

Package: dplyr
Version: 1.1.4
MetricPkgScore: 0.3988960
MetricCovrCoverage: NA
MetricHasNews: 1
MetricRemoteChecks: NA
MetricNewsCurrent: 1
MetricRCmdCheck: NA
MetricExportedNamespace: 0.04611914
MetricHasVignettes: 1
MetricExportHelp: 1
MetricHasWebsite: 1
MetricHasMaintainer: 1
MetricBugsStatus: 0.4333333
MetricSizeCodebase: NA
MetricHasSourceControl: 1
MetricHasBugReportsUrl: 1
MetricDownloads1yr: 0.99194
MetricReverseDependencies: 1
MetricHasExamples: 0.9385965
MetricDependencies: 0.01098694
MetricLicense: NA

```

I think having a dedicated separate function `pkg_metric_export()` allows for more flexibility and easier control over its output without breaking backward compatibility for any tools currently using the package.

This is a starting draft version that basically only formats the current` pkg_score()` output and exports it as dcf. WDYT?


@dgkf @llrs-roche  at this point does this qualify as a minimal example of what we're aiming for? should the values be score or more like assessments?

for example should we have this:

```
MetricHasWebsite: 1
```

or this :

```
MetricHasWebsite: www.websiteOne.com, www.websiteTwo.com
```

dgkf:
This is looking great. I think we could use this as a target for the report to leverage as an input. 

for example should we have this:

> ```
> MetricHasWebsite: 1
> ```
> or this :
> ```
> MetricHasWebsite: www.websiteOne.com, www.websiteTwo.com
> ```

I'm not sure what the best path is here. My first impulse is that the exact urls are already in the DESCRIPTION, so maybe we just want them collapsed into a number for easier programmatic use.

llrs-roche:
Sorry it took me some time to check on this. This looks promising.

When writing two metrics I found that it writes:

```
Package: utils
Version: 4.5.0
MetricDownloads1yr: 0.0000000

Package: dplyr
Version: 1.1.4
MetricDownloads1yr: 0.9919631
```

Note that instead of 0 as it is not an integer due to the dplyr call we get all these decimals. Do you think this could be modified? Or this is `write.dcf` fault (there are [ongoing upgrades](https://github.com/r-devel/r-dev-day/issues/86) on `write.dcf`, so it might be a good time to ask for this feature iff needed) ? 

 I too agree that the number is better. If this is merged with the output of `tools::write_PACAKGES` the websites and other urls might already be there (Depending on the `fields` argument used [I have already asked that R by default it could include all fields but didn't receive any positive feedback]). 

The function is not documented (and it takes a long time to document it given the complex documentation of the "double dispath" mechanism with roxygen2 7.3.2).  But after documenting it worked well.

llrs-roche:
Thanks @eddelbuettel for the info, we'll explore how we can approach this. Maybe just formatting each record and append them to a single file will work.

llrs-roche:
Was playing a bit more with the numeric precision problem:
When exporting just one package instead of 8 decimals we get 15! 
`MetricDependencies: 0.880797077977882` vs `MetricDependencies: 0.88079708`
This might be controlled by `options(digits)` but I am not sure as I don't think I've changed the options between the two runs and I don't see a way to specify as part of `write.dcf` call.

If we export just one package via: `write.dcf(testMany[1, ], file = "out.txt")` the MetricDownloads1yr is a double (with the 8 decimals), if we instead use a single one `write.dcf(testSingle, file = "utils.txt")` we get the integer: `MetricDownloads1yr: 0`. 

We can work around this issue if instead of just `as.character` we use `prettyNum` which is precisely for converting numbers to nice characters: `ops <- options(digits = 8);prettyNum(x$downloads_1yr);options(ops)`, this will also help controlling the decimals used when exporting just one or multiple metrics, and then to read back the object:

```
rd <- read.dcf("file.txt")
rd[, -c(1:2)] <- as.numeric(rd[, -c(1:2)])
rd
```

----

While exploring the inner works I realized `pkg_ref` is deleted instead of keeping the key information: the `source` where this has come from. It would be great if the score could keep that. 
There might be some metrics that change depending on the system, for example some CRAN packages have some alias/help pages that are OS specific...

llrs-roche:
I still see some inconsistencies between the different fields:

```
MetricPkgScore:          0.334186678050244
MetricExportedNamespace: 0.0461191416971477
```

But with the round trip even if the appearance after `apply(read.dcf(file), 1:2, as.numeric)` is different (not sure if we need to address this to not confuse users), I see consistent decimals. I'll add a function to take care of the package and version number on this branch.

PR #387, opened 2025-05-21
add support for exportPattern

### LINKED ISSUE #385: assess_export_help() does not support exportPattern in NAMESPACE
`assess_export_help()` only looks for `export` and `exportMethod` in the `NAMESPACE` file to extract exported objects.

Some `NAMESPACE` files use `exportPattern` instead ([NMF](https://github.com/renozao/NMF/blob/master/NAMESPACE) package for example), which is currently not supported by `assess_export_help()`.

rabiibouhestine:
close #385 

`assess_export_help()` only looks for `export` and `exportMethod` in the `NAMESPACE` file to extract exported objects.

Some `NAMESPACE` files use `exportPattern` instead ([NMF](https://github.com/renozao/NMF/blob/master/NAMESPACE) package for example), which is currently not supported by `assess_export_help()`.

This PR solves this issue, `assess_export_help.pkg_source` now checks for exportPattern lines, extracts regex patterns, and compares them against help_aliases.

### DIFF
--- a/R/assess_export_help.R
+++ b/R/assess_export_help.R
@@ -26,14 +26,29 @@ assess_export_help.pkg_remote <- function(x, ...) {
 #' @export
 assess_export_help.pkg_source <- function(x, ...) {
   pkg_metric_eval(class = "pkg_metric_export_help", {
-    # ignore S3-dispatched methods
+    # Read NAMESPACE
     lines <- readLines(paste0(x$path, "/NAMESPACE"), warn = FALSE)
+
+    # Extract explicit exports
     export_lines <- grep("^export\\(|^exportMethod\\(", lines, value = TRUE)
     exports <- gsub(".*\\(([^)]+)\\).*", "\\1", export_lines)
     exports <- unlist(strsplit(exports, ",\\s*"))
-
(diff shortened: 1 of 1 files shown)


llrs-roche (approved):
Great PR! Very easy to follow and with great comments.
I could verify locally that this works for NMF. I wonder if we can add a test to check that we are now getting the right behavior.

On the bigger picture, this approach is incorrect: there can be R objects that are generated on the package environment that are not documented with the same alias (or vice versa). Perhaps with an example is easier to understand

NAMESPACE:

```
exportPattern(^abc)
```

R/code.R:
```
#' @alias ghost
abcd <- function() "boo"
```

The alias ghost is for function `abcd` which is exported per the pattern but wouldn't be captured (if roxygen2 doesn't add automatically the alias of the assigned object). Some packages that don't use roxygen2 might have done something similar (not easy to quantify/asses) hence this score. 
Notably this happens with some base R functions on the tools package. 

However, packages following current best practices would be correctly identified by this code. So I think it can be merged, at least we will get less false negatives (we identify more exported objects) and it would be difficult that we get some false positives (identifying something as exported that it really isn't).

PR #393, opened 2025-12-05
fix: update to work nicely with upcoming vctrs release

### LINKED ISSUE #392: Issue with `names()` on empty pkg_ref
``` r
names(riskmetric::pkg_ref("utils"))
#>  [1] "bug_reports"        "bug_reports_host"  
#>  [3] "bug_reports_url"    "description"       
#>  [5] "downloads"          "examples"          
#>  [7] "help"               "help_aliases"      
#>  [9] "license"            "maintainer"        
#> [11] "news"               "r_cmd_check"       
#> [13] "release_date"       "remote_checks"     
#> [15] "source_control_url" "vignettes"         
#> [17] "website_urls"       "path"              
#> [19] "source"             "version"           
#> [21] "name"
```

``` r
names(riskmetric::pkg_ref())
#> Error in class(x) <- "environment": "environment" can only be set as the class if the object has this type; found "logical"
```

I think there is something wrong with the empty pkg-ref case that causes this error.

This came up when doing revdeps with https://github.com/r-lib/vctrs/pull/2112 due to the `ptype = pkg_ref()` call here https://github.com/pharmaR/riskmetric/blob/960b4a29d8d0f69b3630b9805b7da39d516c2f67/R/pkg_ref_class_coersion.R#L6 and the fact that `new_list_of()` now calls `vec_ptype(ptype)` on the `ptype` argument to ensure it is a `ptype`, and internally that needs to look at the `names()` of that object, which causes this odd failure.

We plan to release vctrs in early Jan, FYI

dgkf:
Closes #392

A quick fix in order to accommodate upcoming `vctrs` release.

Some whitespace changes introduced due to auto-formatting using `air`. Will highlight key changes inline below.


R/pkg_ref_class.R
```diff
@@ -68,35 +68,59 @@
...
   as_pkg_ref(x, ...)
 }
 
-
+#' @keywords internal
+pkg_ref_ptype <- function() {
+  ptype <- logical()
+  class(ptype) <- c("pkg_ref_ptype", "pkg_ref", class(ptype))
+  ptype
+}
```
dgkf: Instead of having `pkg_ref()` do double duty and using the empty call for our `ptype`, this just leaves that to a separate internal function. We give it a special `_ptype` class suffix to handle some internals that assume a `pkg_ref` is an environment().

PR #394, opened 2025-12-05
fix: upcoming vctrs ptype checks

### LINKED ISSUE #392: Issue with `names()` on empty pkg_ref

dgkf:
Supersedes #393 
Closes #392 

After discovering another issue in a vignette, I realized there's a much simpler solution that I think is more conventional anyways.

### DIFF
--- a/R/pkg_ref_class_coersion.R
+++ b/R/pkg_ref_class_coersion.R
@@ -3,7 +3,7 @@
 #' @export
 as_tibble.pkg_ref <- function(x, ...) {
   as_tibble(vctrs::new_list_of(list(x),
-    ptype = pkg_ref(),
+    ptype = list(),
     class = "list_of_pkg_ref"))
 }
 
--- a/R/utils_memoised.R
+++ b/R/utils_memoised.R
@@ -69,8 +69,6 @@ memoise_bioc_mirrors <- memoise::memoise({
 
 #' @importFrom memoise memoise
 memoise_available_packages <- memoise::memoise({
-  Sys.sleep(20)
-
   function(..., repos = getOption("repos"), .local = getOption("riskmetric.tests")) {
     if (!is.null(.local)) {
       db <- read.csv(
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -45,6 +45,6 @@ Suggests:
     testthat,
     webmockr,
     jsonlite
-RoxygenNote: 7.2.3
+RoxygenNote: 7.3.3
 VignetteBuilder: knitr
 Config/testthat/edition: 3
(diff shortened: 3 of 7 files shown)


R/utils_memoised.R
```diff
@@ -69,8 +69,6 @@ memoise_bioc_mirrors <- memoise::memoise({
 
 #' @importFrom memoise memoise
 memoise_available_packages <- memoise::memoise({
-  Sys.sleep(20)
-
```
dgkf: This one's embarrassing. `git blame` reveals that this is residual from noneother than myself. I don't know exactly why I added it, but I assume it was to test the memoisation. No better time than the present to fix it.

R/pkg_ref_class_coersion.R
```diff
@@ -3,7 +3,7 @@
 #' @export
 as_tibble.pkg_ref <- function(x, ...) {
   as_tibble(vctrs::new_list_of(list(x),
-    ptype = pkg_ref(),
+    ptype = list(),
```
dgkf: `ptype` needs to be a vector type indicating what type of vector-ish object this should be, not a prototype element. This is enforced in the upcoming `vctrs` release and needs to be fixed before early january.

llrs-roche (approved):
Haven't tested locally with the new vctrs but it looks good.

PR #399, opened 2026-02-27
Address devtools revdep check failures in upcoming dev release

dgkf:
For now just drafting changes, will log findings here

### DIFF
--- a/tests/testthat/setup_test_packages.R
+++ b/tests/testthat/setup_test_packages.R
@@ -12,75 +12,102 @@ oldLibs <- .libPaths()
 dir.create(templib)
 
 withr::with_libpaths(c(templib, oldLibs), {
-  devtools::install(
-    file.path(test_path(), "test_packages", "pkgsourcegood"),
-    quiet = TRUE,
-    reload = FALSE
-  )
+  pkg_path <- file.path(test_path(), "test_packages", "pkgsourcegood")
+
+  # build and install
+  pkg_archive <- pkgbuild::build(pkg_path, quiet = TRUE)
+  on.exit(file.remove(pkg_archive))
+  install.packages(pkg_archive, type = "source", repos = NULL, quiet = TRUE)
 
   # a representative "good" quality package from an installed package
   pkg_ref_install_good <- pkg_install("pkgsourcegood")
   assess_install_good <- pkg_assess(pkg_ref_install_good)
   score_install_good <- pkg_score(
     assess_install_good,
-    error_handler = score_error_zero)
+    error_handler = score_error_zero
+  )
 })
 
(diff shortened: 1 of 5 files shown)


tests/testthat/setup_test_packages.R
```diff
@@ -12,75 +12,102 @@ oldLibs <- .libPaths()
...
-    file.path(test_path(), "test_packages", "pkgsourcegood"),
-    quiet = TRUE,
-    reload = FALSE
-  )
+  pkg_path <- file.path(test_path(), "test_packages", "pkgsourcegood")
+
+  # build and install
+  pkg_archive <- pkgbuild::build(pkg_path, quiet = TRUE)
+  on.exit(file.remove(pkg_archive))
+  install.packages(pkg_archive, type = "source", repos = NULL, quiet = TRUE)
```
llrs-roche: I'm not sure how does `on.exit()` play with `withr`

```suggestion
  on.exit(file.remove(pkg_archive), add = TRUE)
  install.packages(pkg_archive, type = "source", repos = NULL, quiet = TRUE)
```
If the installation of the package fails, is there a way to capture and exit early the test? Just to cover all bases. I don't think it fail, but installing packages while testing might be a bit problematic

llrs-roche (approved):
Related to devtools I only see one relevant change from `devtools::install()` to `install.packages()`. I think this will be enough to avoid issues and locally I couldn't detect any issue with devtools 2.5.0. 

I left some comments about minor simplifications of the code too. 

----

While testing I realized that webmockr is not used conditionally as the check failed with an error:

```
Error in `source_dir()`:
! Failed to evaluate './setup_mock_web_requests.R'.
Caused by error in `library()`:
! there is no package called 'webmockr'
```

I would check other packages on Suggests that they are properly used only when the package is available. 

-----

I also got a note (with R 4.5.3) about the documentation:

```
❯ checking Rd files ... NOTE
  checkRd: (-1) pkg_ref.Rd:74-75: Lost braces in \itemize; meant \describe ?
  checkRd: (-1) pkg_ref.Rd:77: Lost braces in \itemize; meant \describe ?
  checkRd: (-1) pkg_ref.Rd:78-80: Lost braces in \itemize; meant \describe ?
  checkRd: (-1) pkg_ref.Rd:81-82: Lost braces in \itemize; meant \describe ?
  checkRd: (-1) pkg_ref.Rd:84-85: Lost braces in \itemize; meant \describe ?
  checkRd: (-1) pkg_ref.Rd:86-87: Lost braces in \itemize; meant \describe ?
  checkRd: (-1) pkg_ref.Rd:88-89: Lost braces in \itemize; meant \describe ?
```

